PR #385, opened 2023-02-20
fix: changed from boolean to type Date (blockedDate and inactiveDate)

admondtamang:
previously, it was boolean.


andrechristikan (approved):

andrechristikan:
My bad... thank you for fixed that.

PR #489, opened 2023-11-01
Feature/v1

phongnv96:


andrechristikan:
@phongnv96
Thank you for you contribute

# [commit 5bde6f7](https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/489/commits/5bde6f7fbecbf8b8d1bce5b9949979ac97e69661) 
In Docker's best practice, as far as I know, it is not necessary to put our env file into an image. 
The .env file can contain sensitive information, such as API keys and database passwords. 

# [commit 63df61c](https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/489/commits/63df61c32a15e32e164fdbc29b813e3243653bf1)
This is the previous code, please update your code.

The main difference is that in SSO with Google, the backend only receives an access token from Google, and if the access token is valid, then the backend will give an access token, just like a login with a password.
If there is any function that requires a Google access token, we can make the endpoint require the token and then send it to Google.
I decided that because the backend does not need to manage the access and refresh from Google.

# [commit fc83c17](https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/489/commits/fc83c17ab1db38a9e949a7395611989d59507f01#diff-521086e3a8b215dcc6bba671d0f9995f7e4fae6e84932f3255f13cb72eaff700)
Actually, I am a newbie in graphql, and I am still learning how to deal with that.
So this will take some time to review.

PR #510, opened 2024-02-26
[CI/CD] Added CD with AWS ECS

kunalrawat425:


andrechristikan:
Thank you @kunalrawat425 
I will check it later

PR #521, opened 2024-08-16
redis fixed (for redis cloud instance)

horizon-dev-group:
The Redis cache works when running in localhost. But for Redis Cloud instance it doesn't work. For it to work the REDIS_USERNAME is required in the env file. So I added that to the required places.

I've also added JOB_ENABLE env variable.


andrechristikan:
Hi @horizon-dev-group ,
Yes, u right. we need REDIS_USERNAME if we want connect to cloud.
Thank you for for u contribution.

But what purpose of added `env. JOB_ENABLE` ?
I didnt see u use that in the project

PR #540, opened 2024-12-19
fix redis cache no default ttl

truongbo17:
Please double check cache time to live with redis, i think this is actually the right option to set default TTL


andrechristikan:
i will check it later, im busy now
anw thx for ur contribution

PR #550, opened 2025-03-28
Fixing a logic problem about the time range

huxiao:
This code does have a logic problem where the time range is reversed, which can result in the API Key being misclassified as inactive and access denied even if the API Key is within the validity period.


andrechristikan:
ahh u are right!
let me merge ur changes

thanks

PR #552, opened 2025-04-21
fix: signup reports no 'user' role

huxiao:
- correct role name to 'individual' as per migration.role.seed.ts


andrechristikan:
merged, thx

PR #554, opened 2025-04-27
fix: Correct logical issues in AuthSharedController.changePassword()

huxiao:
- Prevent resetting password attempt count when old password validation fails.
- Fix password history validation to compare the new password against the old password.
- Avoid direct hash comparison for passwords due to password salting.
- Correct assignment in `HelperDateService.create` when setting `mDate` to start or end of day.
- Improve error message for password reuse validation to provide clearer user guidance.


andrechristikan:
all good, i will merge ur PR
thx

PR #558, opened 2025-05-15
feat(feature-config): feature configuration on database

Gzerox:
Hello, 

I’ve been learning NestJS and TypeScript recently, and this repository has helped me a lot, so I wanted to start giving something back, hopefully this is something someone may find useful.
This PR introduces a simple feature configuration module with support for dynamic flags.

### What it does

1. Loads all feature configurations from the `FeatureConfigs` database collection on app startup.
2. Enables feature-flagging on routes via `@FeatureFlag('feature.name.enabled')`.
3. Allows disabling the cache to always fetch live values.
5. Exposes a admin-method to force a full reload of the configuration from the database.

### Motivation

In my current project, we manage a large number of configuration flags via environment variables. 
This comes with a few limitations:
- Updating any value requires restarting the application.
- For automated environments, even minor changes requires a full redeployment.
- It’s difficult to manage or document what each env var does.
  
With this module, I wanted to:
- Enable runtime configuration updates without restarts.
- Support a simple feature flagging
- Add caching control and live reloading options.

------------------------------

#### Protect an Endpoint Based on a Feature Flag

You can guard routes using the `@FeatureFlag()` decorator:
```javascript
@FeatureFlag('feature.enabled')
@Get("specialfeature")
async test(): Promise<IResponse<any>> {
	return {
		data: "hello world",
	};
}
```

If your `FeatureConfigs` collection **does not contain** an entry with the `key=feature.enabled`, the feature will default to false and return a 6000 status:

```json
{
  "statusCode": 6000,
  "message": "This feature is currently disabled",
  "_metadata": {
    "language": "en",
    "timestamp": 1747329393218,
    "timezone": "Asia/Jakarta",
    "path": "/api/v1/admin/features/specialfeature",
    "version": "1",
    "repoVersion": "7.4.0"
  }
}
```

The same response will be returned if a corresponding entry exists in the database and is explicitly set to `false`.

#### Reload configuration from database

By default, feature configurations are cached on application startup. That means if you update a value in the FeatureConfigs collection (changing `service.endpoint.timeout`), the change **won’t be picked up immediately**:

```javascript
export const SERVICE_ENDPOINT_TIMEOUT = 'service.endpoint.timeout';

export class YourBusinessLogicService {  
  constructor(private readonly config: FeatureConfigService) {  
  }
	async performStuff(...) {  
	  const serviceEndpointUri = await this.config.get<string>(SERVICE_ENDPOINT_TIMEOUT);
	}
}
```
To refresh the configuration **without restarting the application**, you can trigger a reload via the admin endpoint:
```
POST {{host}}/api/admin/features/reload
Authorization: Bearer xxxxxx
x-api-key: yyyyyy
```

This will clear the cache and reload all feature configs from the database.

#### Fetch a Feature Configuration
You can retrieve feature configurations within your service logic using the FeatureConfigService. Here’s a basic example:
```javascript

export const SERVICE_ENDPOINT = 'service.endpoint.url';
export const SERVICE_ENDPOINT_TIMEOUT = 'service.endpoint.timeout';
export const SERVICE_ENDPOINT_SOMETHING_ESLE = 'service.endpoint.something.esle';

export class YourBusinessLogicService {  
  constructor(private readonly config: FeatureConfigService) {  
  }

	async performStuff(...) {  
	  const serviceEndpointUri = await this.config.get<string>(SERVICE_ENDPOINT);
	  const serviceEndpointTimeoutMs = await this.config.get<number>(SERVICE_ENDPOINT_TIMEOUT, 3000);
	  // I want this one to be fetched from DB everytime.
	  const serviceEndpointSomethingElse = await this.config.get<string>(SERVICE_ENDPOINT_TIMEOUT, null, true);
  }
}
```
> This  could be extended or improved further, but it’s enough as starting point.

## If you don't like/need it
This feature is intended to be very simple & basic. 
If it doesn’t suit your needs, you can easily disable it by commenting out the following line in `common.module.ts`:
```javascript
@Module({  
    controllers: [],  
    providers: [],  
    imports: [
	    //... all your imports
	    
	    // Comment the following line to DISABLE 
		FeatureConfigModule.forRoot(),  
    ],  
})
export class CommonModule {}
```

And then remove the controller registration from `routes.admin.module.ts`:

```javascript
@Module({  
    controllers: [  
	    // Comment the following line to DISABLE 
        FeatureConfigAdminController  
    ],  
    //..... 
})  
export class RoutesAdminModule {}
```

That's it.


yonashalefom:
This is really nice!

andrechristikan:
Hi @Gzerox 
This is a huge feat, currently im super busy
I will read it carefully once i have some time and also i will give a feedback too

Btw this is outstanding feat! 
Lets make this feat together 🙌

src/common/features/bases/base-config.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  async deleteMany(
+    find: Record<string, any>,
+    options?: IDatabaseDeleteManyOptions,
+  ): Promise<boolean> {
+    await this.settingRepository.deleteMany(find, options);
+
+    return true;
+  }
```
blackmouse572: Why don't we update `cache` here ? It could possible user delete some of value in cache

src/common/features/bases/base-config.service.ts
blackmouse572, line 1: We can add generic `T` at the top to remove redundant 
```tsx
export abstract class AppBaseConfigService<
  TValue,
  TEntity extends AppBaseConfigEntityBase<TValue> = AppBaseConfigEntityBase<TValue>,
  TDoc extends IDatabaseDocument<TEntity> = IDatabaseDocument<TEntity>,
  TRepo extends AppBaseConfigRepositoryBase<TEntity, TDoc> = AppBaseConfigRepositoryBase<TEntity, TDoc>
> implements OnModuleInit {
  // ...
  async get(key: string, fallback?: TValue, forceReload = false): Promise<TValue | undefined> { ... }

  async set(key: string, value: TValue): Promise<void> { ... }
}
```

and use
```tsx
interface MailConfig {
  host: string;
  port: number;
  user: string;
  pass: string;
}

class MailConfigEntity extends AppBaseConfigEntityBase<MailConfig> {}
class MailConfigService extends AppBaseConfigService<
  MailConfig,
  MailConfigEntity,
  IDatabaseDocument<MailConfigEntity>,
  AppBaseConfigRepositoryBase<MailConfigEntity, IDatabaseDocument<MailConfigEntity>>
> {}
```

src/common/features/controllers/feature-config.admin.controller.ts
blackmouse572, line 1: I think you are missing doc for this controller ?

blackmouse572:
Excellent work for new TS comer 👍🏻

src/common/features/bases/base-config.service.ts
Gzerox: I thought a bit about this, generally, I don't think its a good idea to automatically refresh the `in-memory cache` immediately after a change is being triggered from the service.

I like to separate the `change the fields on database` and the `apply/publish` the changes to the running application.
This just to make sure the reload is very intentional.

While im typing this, I noticed the current behaviour of the code (such as `set` ) is not reflecting perfectly my thoughts, I'll do some changes to reflect that.

andrechristikan:
hi @Gzerox 
great job! thanks for ur contribution
this is a good feat, but we need to some improvement

let me give my feedback

1. let decide about the module name. is `feature config` common ?
because i never hear it 😆

2. why u created those are files, and why dont u just put it directly under `repository/*` dir ?
or do u think this module can extends to other ? i dont understand
- `base-config.entity`
- `base-config.repository.ts`
- `base-config.service`

3. do u think we can merge this feat with setting module ? 

4. dont use `map` for cache `protected cache = new Map<string, any>();`.
this project already use redis, i think would be better if u can use it.
my reason is... think about, if we want to make a project thats need to run multiple instances.
this feature should be able to do that

5. why dont u use pagination in `/` and `/cache` endpoint ? and what if we have so many feature-configs ?

6. some files naming are incorrect

Gzerox:
Hey @andrechristikan 

> let decide about the module name. is feature config common ?
> because i never hear it 😆
>
> do u think we can merge this feat with setting module ?

This feat started with the name of `Feature Flag`.

Naming… always the hardest part !
Jokes aside, I’ve generally seen the `Settings module` used for **user-specific preferences**, while `Configuration` feels more like **system-level behavior**.
If the existing `Settings module` isn’t meant to be tied to a specific user, then yeah, I think this could be a solid entry point for the feature.

> why u created those are files, and why dont u just put it directly under repository/* dir ?

I followed the same structure used in `./common/database/bases`, just for consistency.

> or do u think this module can extends to other ? i dont understand

Honestly, I don’t fully remember the reasoning behind it 😅
But I do recall considering things like:
1. A common feature configurations - this includes `system configuration` and `user configuration`
2. Having the option to specify `feature configuration` for specific user
3. Having the option to specify `feature configuration` for user group/role/policy.

This things would allow us to achieve stuff like:
1. Controlling the feature configs for the `normal user`
2. Enable/Disable/Limiting/Extending features per user or group/role/policy (such as tester,qa, preview).

In those cases, we might eventually need to extend AppBaseConfigEntityBase to include extra fields, etc.

That said, I totally get that for now these extra classes might be overkill—no need to over-complicate things (something I’m occasionally guilty of 😅).

> dont use map for cache protected cache = new Map<string, any>();.
> this project already use redis, i think would be better if u can use it.
> my reason is... think about, if we want to make a project thats need to run multiple instances.
> this feature should be able to do that

That makes total sense. 
I’m not using Redis in my own setup yet, so I went for something quick and local, but I get your point, I’ll refactor it.

> why dont u use pagination in / and /cache endpoint ? and what if we have so many feature-configs ?

Ah, good catch.
I’m not using those endpoints myself so I totally missed it. 
I’ll go ahead and add pagination.

Thanks again for the feedback
I’ll work on the changes over the next few days and request another review once they’re ready.
Happy to get constructive input, so feel free to point things out as we go. 

Just bear with me if I miss something here and there 🙂

andrechristikan:
hi @Gzerox 
i really appreciate the effort you put into this

i agree that naming is always the hardest part! 😆 
Since our current setting module isn't tied to any specific feature yet (it only has one endpoint showing core project settings)
I think we could merge this feature config functionality into the existing setting module. what do you think?

and for the initial implementation, how about we focus on feature-level config only? 
this would avoid over complexity while still providing the core functionality
we can always extend it to user/role-specific configs later if needed

if we integrate this with existing modules in the project, 
I think we could start with something like (feel free to add it):
- `auth.social.google.enabled`
- `auth.social.apple.enabled`

let me know your thoughts! really excited to see this feature in the project 🚀

Gzerox:
oh boy, I did screw up a bit with the git-rebase and conflict resolutions :D

Gzerox:
Do you see anything else that could be improved or added?

I agree that starting with social login makes the most sense.
Would you prefer to include that in this PR, or should we merge this first and handle it separately?

In the meantime, this can already be used to start experimenting with new features and configurations.

andrechristikan:
hi @Gzerox 
first of all, let me merge ur PR into new branch
then i will make some changes regarding the feat

i want to change the module name to `SettingFeature` because i think its more fit
plus let me change a bit before then u can review it too

andrechristikan:
@Gzerox 
here u go #568

PR #560, opened 2025-05-20
feat: auto update the kids in .env when running yarn generate:keys

huxiao:
- Copy .env from .env.example if missing
- Update the KIDs in .env according to jwks.json


andrechristikan:
thats awesome! 

let me merge it, thanks

PR #568, opened 2025-06-11
Feat/setting feature

andrechristikan:
1. change module name to setting feature
2. do serialize and deserialize before save record into redis
3. using only 1 type of value for setting `{ enabled: boolean, [key: string]: any}`

todo: 
- [x] 1 more api to update the setting
- [x] implement the setting at login with google with `auth.social.google.enabled` as key
- [x] implement the setting at login with apple with `auth.social.apple.enabled` as key


Gzerox:
Thanks @andrechristikan for refactoring and adjusting it a bit :)

From what I understand, this change forces every setting to follow a “complex” object format, like:
```
key = auth.social.google
value = {
  enabled: true,
  setting1: "value1",
  setting2: "value2"
}
```

This means we’d no longer support the simpler key/value format, such as:
```
key = auth.social.google.enabled
value = true
```
or 
```
key = auth.social.google.setting1
value = "value1"
```

In previous implementation, both formats were supported, even if it wasn’t perfect and maybe too flexible. 
I understand that supporting both adds complexity, so if we’re intentionally choosing to support only the complex format for consistency and easier management, i am totally okay with that.

----

That said, just want to make sure we’re all aligned, because this also affects how we retrieve and use settings in code. 

For example:

```javascript
// won't work anymore, since the key ".enabled" doesn't exist on its own
const setting = await this.settingFeatureService.findOneByKey('auth.social.google.enabled');
const isEnabled = setting.value.enabled;

// Instead we have to do:
const setting = await this.settingFeatureService.findOneByKey('auth.social.google');
const isEnabled = setting.value.enabled;
```
NOTE: `findOneByKey` does not exists as method, is just for example purposes :D

So the change might require some naming adjustments, like having one key per feature domain (auth.social.google) instead of multiple fine-grained keys.

Let me know if that’s the intended direction !

Gzerox:
after some thought I believe its better if we stick with the current `complex` type.
There's no real benefit of managing and maintaining multiple modes (simple & complex).

Gzerox:
Im having hard time understanding how `value` :

```javascript
    @DatabaseProp({
        type: {
            enabled: { type: Boolean, default: false },
            additionalProperties: {
                type: Schema.Types.Mixed,
            },
        },
        required: true,
    })
    value: SettingJson;
```

will work with `SettingJson`

```javascript
export interface SettingJson {
    enabled: boolean;
    [key: string]: SettingValue;
}

export type SettingValue =
    | boolean
    | string
    | number
    | SettingValue[]
    | SettingJson;
```

From the schema, I get we should add the setting values under `additionalProperties`, right ?

```json
{ 
    "enabled": true,
    "additionalProperties": {
        "myconfig1": "value1",
        "myconfig2": "value2",
    }
}
```

Im not sure this will work, im expecting to get some type assignment issue, am I missing something ?

andrechristikan:
hi @Gzerox 
> after some thought I believe its better if we stick with the current complex type.
There's no real benefit of managing and maintaining multiple modes (simple & complex).

Here are my concerns about maintaining multiple modes (simple & complex):

1. **Maintenance overhead** - Supporting two different data structures will increase complexity in the long run
2. **Feature consistency** - Having a single, well-defined approach ensures consistent behavior across the application  
3. **Future scalability** - The complex type gives us more flexibility for future enhancements without breaking changes

Since we already have the logic implemented for the complex approach, I think it makes sense to go with that. It provides better structure and extensibility.

> So the change might require some naming adjustments, like having one key per feature domain (auth.social.google) instead of multiple fine-grained keys.

Yes, that's exactly the intended direction. The hierarchical approach with domain-based keys makes more sense for the complex type structure.

So we'll have:
- `auth.social.google` → `{ enabled: true, ... }`
- `auth.social.apple` → `{ enabled: false, ... }`

This gives us better grouping and reduces the number of individual setting entries. It also makes the configuration more manageable when we have features with multiple related properties.

andrechristikan:
> Im having hard time understanding how `value` :
> 
> ```js
>     @DatabaseProp({
>         type: {
>             enabled: { type: Boolean, default: false },
>             additionalProperties: {
>                 type: Schema.Types.Mixed,
>             },
>         },
>         required: true,
>     })
>     value: SettingJson;
> ```
> 
> will work with `SettingJson`
> 
> ```js
> export interface SettingJson {
>     enabled: boolean;
>     [key: string]: SettingValue;
> }
> 
> export type SettingValue =
>     | boolean
>     | string
>     | number
>     | SettingValue[]
>     | SettingJson;
> ```
> 
> From the schema, I get we should add the setting values under `additionalProperties`, right ?
> 
> ```json
> { 
>     "enabled": true,
>     "additionalProperties": {
>         "myconfig1": "value1",
>         "myconfig2": "value2",
>     }
> }
> ```
> 
> Im not sure this will work, im expecting to get some type assignment issue, am I missing something ?

ahhh you're absolutely right! 
there's a mismatch between the MongoDB schema and TypeScript interface.

the current schema structure won't work as intended.
we need to align them. I think we should go with the **flat structure** approach:

```typescript
@DatabaseProp({
    type: Schema.Types.Mixed,  // Just use Mixed directly
    required: true,
})
value: SettingJson;
```

andrechristikan:
Let me know if you need clarification on either approach @Gzerox 

And if all clear, could you please continue this PR?
1. Fix the schema mismatch (remove `additionalProperties` nesting)
2. Implement domain-based keys (use `auth.social.google` instead of `auth.social.google.enabled`)

Gzerox:
there you go: #569 @andrechristikan 

hopefully we are near the finish line :D

andrechristikan:
@Gzerox cool man, thanks
dont worry i will take the rest

once im done, i will tag u

andrechristikan:
hi @Gzerox 

here are my updates

### Remove create and delete endpoints
I don't think we need create or delete endpoints for settings because there's no point for users to do that. If there's a new setting, we as developers need to adjust the code anyway.

### Change cache strategy
Instead of loading all settings at startup and caching via /reload API, it would be better to use lazy loading. We only cache the setting once it's accessed. When we update a setting, the setting will be automatically deleted from cache. We can also /flush the cache via endpoint.

Benefits:
- Faster startup time (no preloading)
- Memory efficient (only cache used settings)

how do u think?
let me know so we can merge it and add it to the docs.

Gzerox:
Hey @andrechristikan,

Hard to disagree with you related to Lazy Load approach !, makes sense.

As of now, like you mentioned,I don't see how someone would use Create/Delete method in any way, I just added those `just-in-case` ^^ .

andrechristikan:
Hi @Gzerox 
The documentation is complete. 
I will merge this feature into the development branch.

Thank you for your contribution!

PR #571, opened 2025-06-15
naming-convention: renaming a few dtos

Gzerox:
nothing really to describe.

I just noticed some file where not matching the naming convention used.


andrechristikan:
hi @Gzerox 
u're awsome!

let me merge it, thanks

PR #573, opened 2025-06-18
fix: increase-password-attempt pipeline invalid syntax

Gzerox:
Hello,

Im not sure this is the desired way to fix this, however:

### The issue:

```javascript
async increasePasswordAttempt(
    repository: UserDoc,
    options?: IDatabaseUpdateOptions
): Promise<UserDoc> {
    return this.userRepository.updateRaw(
        { _id: repository._id },
        this.databaseService.aggregateIncrement('passwordAttempt', 1),
        options
    );
}
```

While running it as-is:
```
    message: "Data must be an array"
    err: {
      "message": "Data must be an array",
      "stack":
          Error: Data must be an array
              at UserRepository.updateRaw (.../src/common/database/bases/database.object-id.repository.ts:491:13)
              at UserService.increasePasswordAttempt (.../src/modules/user/services/user.service.ts:413:32)
              at AuthPublicController.loginWithCredential (/.../src/modules/auth/controllers/auth.public.controller.ts:97:37)
              at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    }
```

Even if you wrap it as array:

```javascript
async increasePasswordAttempt(
    repository: UserDoc,
    options?: IDatabaseUpdateOptions
): Promise<UserDoc> {
    return this.userRepository.updateRaw(
        { _id: repository._id },
        [ this.databaseService.aggregateIncrement('passwordAttempt', 1) ],
        options
    );
}
```

```
Invalid update pipeline operator: "$inc"
```

### DIFF
--- a/src/modules/user/services/user.service.ts
+++ b/src/modules/user/services/user.service.ts
@@ -46,7 +46,6 @@ import { UserUpdateStatusRequestDto } from '@modules/user/dtos/request/user.upda
 import { DatabaseHelperQueryContain } from '@common/database/decorators/database.decorator';
 import { UserUploadPhotoRequestDto } from '@modules/user/dtos/request/user.upload-photo.request.dto';
 import { UserCensorResponseDto } from '@modules/user/dtos/response/user.censor.response.dto';
-import { DatabaseService } from '@common/database/services/database.service';
 
 @Injectable()
 export class UserService implements IUserService {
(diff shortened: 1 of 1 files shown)


andrechristikan:
Thank you for finding this issue, and technically your solution is right.
But it seems that doesn't hit the root cause. After checking, the root cause is in `database.repository.ts`. 

There's a slight logic error there. 
If you want to see it, you can check the main branch. 
I just updated it.

PR #577, opened 2025-06-21
feat(terms-policy): added module for managing terms and policies

Gzerox:
As always im seeking advice from you on how to improve/restructure and make it better :)

## Introduction

When building website/service, we often have to request the user to accept a series of policies/terms, some of them maybe optional and some other required in order to use the service/website, this could easily be affected by country and laws.

## Feature Overview 

The feature aims to streamline the process of accepting policies and terms that are essential for using the platform. 

Without going too deep in the topic, we can categorize the terms & policies:

| Name             | Note                                                                         |
| ---------------- | ---------------------------------------------------------------------------- |
| Terms of Service | Rules for using the platform (e.g. rights, restrictions, termination).       |
| Privacy          | Explains how personal data is collected, used, and stored (GDPR, CCPA, etc). |
| Cookie           | Informs about tracking cookies; consent often required (especially in EU).   |
| Refund & Return  | Sets rules for refunds or returns (typically for ecommerce).                 |
| Community        | Behavioral expectations in social or user-generated content platforms.       |
| Marketing        | Allows sending promotional emails; opt-in usually required.                  |
| OTHER            | There may be many other, which you can easily add to the application         |

Excerpt of `ENUM_TERMS_POLICY_TYPE`:

```javascript
export enum ENUM_TERMS_POLICY_TYPE {  
  TERMS = 'terms',  
  PRIVACY = 'privacy',  
  COOKIES = 'cookies',  
  MARKETING = 'marketing',  
  COMMERCIAL = 'commercial',  
}
```

For every `terms_type`, `lang` and `version` we can only have a unique entry (unique composite index)

## Versioning & Tracking
This was not something i was really thinking about implementing, but i saw in the readme that the versioning would have been appreciated so i decided to spent some time thinking how could this be done.

Im confident this solution could be vastly improved, this is what i came up with my 2 brain cells :D

### Terms and Policy versioning tracker
For every new update we will take care of increasing a integer on the collection `TermsPolicy.version`.
Whenever a term is published it should not be possible to update it, a new version need to be created.
### User acceptance tracker
Simple collection to keep track of the user acceptance, the version, and the type of terms/policies being accepted.

```javascript
export class TermsPolicyAcceptanceEntity extends DatabaseUUIDEntityBase {  
    @DatabaseProp({  
        type: String,  
        ref: UserEntity.name,  
        required: true,  
        index: true,  
    })  
    user: string;  
  
    @DatabaseProp({  
        type: String,  
        ref: TermsPolicyEntity.name,  
        required: true,  
        index: true,  
    })  
    policy: string;  
  
    @DatabaseProp({ required: true, type: Date })  
    acceptedAt: Date;  
}
```

## Endpoints

| Endpoint                          | Description                                                               |
| --------------------------------- | ------------------------------------------------------------------------- |
| `public/terms-policy/:lang`       | return the list of latest published version policies for a given language |
| `public/terms-policy/:lang/:type` | return the most recent published terms/policy for given type and language |
| `user/terms-policy/pending`       | Logged-in user receives list of "not-(yet)-accepted" policies.            |
| `user/terms-policy/accepted`      | Logged-in user receives back the list of accepted policies.               |
| `user/terms-policy/accept`        | Logged-in user accept a given policyId (passed as JSON Body)              |
| `admin/terms-policy/xxx`          | Admin Endpoints                                                           |
## Consent during signup

A new addition to the signup DTO `AuthSignUpRequestDto`

```javascript
// AuthSignUpRequestDto
@IsObject()  
@IsNotEmptyObject()  
@ValidateNested()  
@Type(() => TermsPolicySignupRequestDto)  
legal: TermsPolicySignupRequestDto

// TermsPolicySignupRequestDto
@ApiProperty({  
    description:  
        'Indicates whether the user accepted the Terms of Service. Must be true.',  
    example: true,  
})  
@RequireUserAgreement()  
@IsBoolean()  
[ENUM_TERMS_POLICY_TYPE.TERMS]: boolean;  
  
@ApiProperty({  
    description:  
        'Indicates whether the user accepted the Privacy Policy. Must be true.',  
    example: true,  
})  
@IsBoolean()  
@RequireUserAgreement()  
[ENUM_TERMS_POLICY_TYPE.PRIVACY]: boolean;

// OTHER
```

Which will require the user to agree to a set of terms/policies defined in `TermsPolicySignupRequestDto`

## Acceptance Tracker
Simple collection to keep track of the user acceptance, the version, and the type of terms/policies being accepted.

## Method Decorators
You may find yourself wanting to block access to a given endpoint to only users which accepted a given terms/policy.

In such case you could annotate your method with `@TermsPolicyAccepted(ENUM_TERMS_POLICY_TYPE.PRIVACY)`, such as:

```javascript
@Get('/protected/privacy')  
@TermsPolicyAccepted(ENUM_TERMS_POLICY_TYPE.PRIVACY)  
@UserProtected()  
@AuthJwtAccessProtected()  
async getProtectedByPrivacy() {  
  return {  
    message: 'You have accepted the policy of privacy',  
  };  
}
```

The user **needs to be logged-in**.

If the users accepted the currently active Privacy Terms, then it is allowed to access the endpoint.

If the user didn't accepted the latest `Privacy Policies` :
```json
{
  "statusCode": 6100,
  "message": "You need to accept privacy policy before you can use this feature.",
  "_metadata": {
    "language": "it",
    "timestamp": 1748029183872,
    "timezone": "Europe/Rome",
    "path": "/api/auth/terms-policy/protected/cookies",
    "version": "1",
    "repoVersion": "0.2.1"
  }
}
```

NOTE: be careful about the decorators order.

### Different language, different policies
Whenever the user accepts `terms_policy=cookies` with `lang=en`, if at later point decide to change the language, the new policy `terms_policy=cookies && lang=jp` will result not accepted (even if is the same, in a different language.)

### Handling the type-requirement at DB level
For flexiblity purposes we may need to have a `required` flag on the collection level instead of in the DTO.


andrechristikan:
Hi @Gzerox 
Thanks for this comprehensive PR!

This is a really interesting feature
the terms & policies management system looks well thought out, especially the versioning approach and the decorator for access control.

This feature complements the authentication and authorization capabilities
and makes more production-ready for real-world applications 🔥

I've checked some parts already, just need to review the `/modules/*` folder in detail. Give me some time to digest all the implementation and I'll provide proper feedback.

### A few quick questions
1. `AppCountryMiddleware`
I noticed this middleware was added but I'm not seeing the clear connection to the terms & policies feature. Could you explain what role this middleware plays and how it integrates with the overall functionality?

2. In repository: why the change to `import { Schema as MongooseSchema } from 'mongoose'`;
I see you've aliased the Schema import - is there a specific reason for this? Usually we just use `import { Schema } from 'mongoose'` unless there's a naming conflict. Just want to understand the reasoning behind this change.

### Code feedback
#### Language consistency
`RequireBooleanConstraint` -> should use camelCase following our naming conventions, so `requireBooleanConstraint`

#### Migration naming
`MigrationTermSeed` -> should be more descriptive, suggest changing to `MigrationTermPolicySeed` to be clearer about what it's seeding

#### Seed content approach
  1. For the seed content, instead of storing the actual content as strings in the database, consider uploading the content as HTML files to S3 and storing the S3 URLs in the seed. This approach is more scalable and allows for richer content formatting.
  2. I noticed you're planning to seed multiple versions - what's the reasoning behind this? Are you trying to simulate version history or is there a specific use case that requires multiple initial versions?

Will get back with more detailed review once I go through the modules folder. Thanks for the solid contribution! 🚀

Gzerox:
Hey @andrechristikan !

Thanks a lot for the feedback ! 
The PR is still in draft exactly for this reason: to get your, and others,input on how to improve the approach before finalizing it.

Lets go point by point.

> AppCountryMiddleware
> I noticed this middleware was added but I'm not seeing the clear connection to the terms & policies feature. Could you explain what role this middleware plays and how it integrates with the overall functionality?

The middleware is meant to lay the foundation for multi-country support, where each country could have different legal requirements (slightly or completely).

The scope of the PR (for now), handling a single-country setup, defined via the `APP_COUNTRY` environment variable. 
But the idea is to future-proof the system to support two main scenarios:

### Current: Multi-Instance, Multi-Country
Assume we deploy the same API in different regions, sharing the same database:
1. API deployed in UK => `APP_COUNTRY=UK`
2. API deployed in Germany => `APP_COUNTRY=DE`

Each instance will return the legal terms & policies relevant to its configured country. (mainly because we will add a WHERE condition on the service-layer to filter by `country` field)

### Future: Single-Instance, Multi-Country (Multi-Domain)
In the future, we might want a single API instance to serve multiple domains.
In that case, the middleware would detect the incoming domain (or similar) and resolve the country dynamically:

1. `api.ack.co.uk` → resolves to country=UK
2. `api.ack.de` → resolves to country=DE
3. `api.ack.es` → resolves to country=ES
You get the idea.

So in summary: while right now the middleware only reads from the environment, it’s designed to scale to more advanced, domain-based multi-country support later on.

---

> In repository: why the change to import { Schema as MongooseSchema } from 'mongoose';
> I see you've aliased the Schema import - is there a specific reason for this? Usually we just use import { Schema } from 'mongoose' unless there's a naming conflict. Just want to understand the reasoning behind this change.

Honestly don’t recall doing that on purpose, I'll double check later and let you know.

> Language consistency
> RequireBooleanConstraint -> should use camelCase following our naming conventions, so requireBooleanConstraint
> 
> Migration naming
> MigrationTermSeed -> should be more descriptive, suggest changing to MigrationTermPolicySeed to be clearer about what it's seeding

Agree, I was even thinking to rename the whole thing into something like `LegalPolicy` or `LegalTerm`, `LegalText`, but I was hoping someone would step in with better naming ideas than mine


### Seed content approach
> For the seed content, instead of storing the actual content as strings in the database, consider uploading the content as HTML files to S3 and storing the S3 URLs in the seed. 
> This approach is more scalable and allows for richer content formatting.

Oh, I wasn’t really thinking of doing it that way, but I’m open to it.

I do have a couple of questions on how you’d see it working in practice.
how the API response would look like ?
Would the endpoint still return the full content like this?

```json
{
  "message": "Your terms and policies",
  "data": "VERY LONG CONTENT READ FROM S3 FILE"
}
```

or more like: 

```json
{
  "message": "Your terms and policies",
  "data": {
    "path": "https://s3.amazonaws.com/yourinfo.html"
  }
}
```

Also, storing content in S3 means introducing additional AWS dependencies, so we need to think about IAM policies, S3 permissions.

Would like to hear your thoughts on how you’d go with this integration.

---

> I noticed you're planning to seed multiple versions - what's the reasoning behind this? 
> Are you trying to simulate version history or is there a specific use case that requires multiple initial versions?

You got it, just wanted to see how the code behave in different language/country/version scenarios, I see no reason to have a initial seed with multiple entry version.
But yep, this shouldn’t have made it into the committed seed, I’ll clean that up.

Gzerox:
> In repository: why the change to import { Schema as MongooseSchema } from 'mongoose';
> I see you've aliased the Schema import - is there a specific reason for this? Usually we just use import { Schema } from 'mongoose' unless there's a naming conflict. Just want to understand the reasoning behind this change.

I have an answer for this.

Just to make sure we are on the same page:

I was not able to find a specific match for the import you were referring to `import { Schema } from 'mongoose'` (other than in `src/modules/setting/repository/entities/setting-feature.entity.ts`

I believe the change you are referring to is related to `src/common/database/decorators/database.decorator.ts`:


```javascript
// Before
import { SchemaType as MongooseSchema } from 'mongoose';

// After my change
import { Schema as MongooseSchema } from 'mongoose';
```

The reason for this change is because while using: `import { SchemaType as MongooseSchema } from 'mongoose';`:
The composite index on my entities are complaining:

```javascript
// terms-policy.acceptance.entity.ts
TermsPolicyAcceptanceSchema.index({ user: 1, type: 1, country: 1, version: 1 }, { unique: true });

// terms-policy.entity.ts
TermsPolicySchema.index({ type: 1, language: 1, country: 1, version: 1 }, { unique: true });
```
Error:
```
src/modules/terms-policy/repository/entities/terms-policy.acceptance.entity.ts:64:81 - error TS2554: Expected 1 arguments, but got 2.

64 TermsPolicyAcceptanceSchema.index({ user: 1, type: 1, country: 1, version: 1 }, { unique: true });
src/modules/terms-policy/repository/entities/terms-policy.entity.ts:59:75 - error TS2554: Expected 1 arguments, but got 2.

59 TermsPolicySchema.index({ type: 1, language: 1, country: 1, version: 1 }, { unique: true });

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

andrechristikan:
Hi @Gzerox,
I'll categorize this into 6 topics to make our discussion easier:

## AppCountryMiddleware
I understand the need for country-based policies. Then let's create a country middleware similar to the language middleware approach.

It should be controllable via headers, giving frontend flexibility to choose what they need. Since language uses `x-custom-lang`, maybe we can use `x-country` or `x-custom-country` for country? Feel free to suggest better naming.

For reference, check `src/middleware/language.middleware.ts` and `src/middleware/language.middleware.spec.ts`

## Naming Convention
Usually, naming starts with the module name followed by the feature name.

For example: if the module is migration and the purpose is to seed term policies, it should be MigrationTermPolicySeed.

Without this pattern, it becomes difficult to find features when the codebase grows larger.

## Content Storage Approach
Typically, we only store content links in the database and return those links to the frontend. 
The frontend is then responsible for fetching and rendering the content.

This approach offers several benefits:
- Frontend flexibility to modify content without backend changes
- No need to worry about content updates on backend side
- Avoids server data exchange limits (length/size restrictions)
- Prevents arguments with DevOps about payload sizes 😄

Example response:
```json
{
  "message": "Your terms and policies",
  "data": {
    "content": "https://s3.amazonaws.com/yourinfo.html",
    ...
    ...
  }
}
```

## Regarding S3 policies 
don't worry about that. It's outside this project's scope. Assume developers know how to use S3 and configure it properly. Without this scope limitation, our feature would become overly complex.

## Seed Values
If you want to demo versioning, it's better to do it manually. Seeds should be focused on initial data only.

## Schema vs SchemaType
After researching this, there was a mistake. It should be `Schema` not `SchemaType`. And yes, i agree with you.

src/modules/auth/dtos/request/auth.sign-up.request.dto.ts
```diff
@@ -22,4 +24,10 @@ export class AuthSignUpRequestDto extends OmitType(UserCreateRequestDto, [
     @MinLength(8)
     @MaxLength(50)
     password: string;
+
+    @IsObject()
+    @IsNotEmptyObject()
+    @ValidateNested()
+    @Type(() => TermsPolicySignupRequestDto)
```
andrechristikan: Don't forget to use `@IsNotEmpty()` and add @ApiProperty(). Also, what do you think about using `termPolicies` for the field name? I think it might be better since it represents a collection, but I'm open to your thoughts on this.

src/modules/auth/controllers/auth.public.controller.ts
```diff
@@ -394,13 +400,20 @@ export class AuthPublicController {
                 {
                     email,
                     name,
-                    password: passwordString,
```
andrechristikan, line 397: I think you made a mistake here

`password: passwordString` should not be removed
It will cause a lint error.

src/modules/terms-policy/dtos/request/terms-policy.signup.request.dto.ts
andrechristikan, line 1: Actually, I don't really like this pattern because it doesn't look clean. 
Maybe we can just use explicit field names instead?

like this
```
terms: boolean
privacy: boolean
cookies: boolean
marketing: boolean
```

andrechristikan:
Hi @Gzerox,

Before we go too far, how to change the module name using singular approach to be `term-policy` instead of `terms-policy`?

src/modules/auth/dtos/request/auth.sign-up.request.dto.ts
Gzerox: I think you just tagged someone involuntarily ahahah :D

src/modules/terms-policy/dtos/request/terms-policy.signup.request.dto.ts
Gzerox: I thought you were going to say something similar.
I went with that approach cause I wanted to make sure all the keys are kept in sync when refactoring, useful to find reference "where is this used", I'll try a different approach, lets see if its better.

src/modules/auth/dtos/request/auth.sign-up.request.dto.ts
Gzerox: Technically, the `@IsNotEmptyObject` should already cover `@IsNotEmpty` evaluations, what am I missing ?
Agree with the plural in this case !

src/modules/auth/controllers/auth.public.controller.ts
Gzerox: “It works on my machine!” jokes aside...  

I know this might usually trigger a lint warning.

```javascript
// Before:
{ email, name, country }: AuthSignUpRequestDto,

// After:
{ email, name, country }: Omit<AuthSignUpRequestDto, 'termPolicies' | 'password'>,
```

Im aware its becoming ugly, however it the only quick-way i found to exclude specific field from the type.

We are using neither `termPolicies`, `password` in the service-method, but we are required to pass to the function due to `AuthSignUpRequestDto`.

Gzerox:
thanks for the feedbacks @andrechristikan ,

I'll implement the code-changes based on your feedbacks, however, there's one left which im not quite clear: **Content Storage Approach**.

I understand the reasoning behind storing terms/policies as static HTML on S3, and it makes sense in terms of scalability and rendering flexibility. 
However, I’m trying to better picture how the full CRUD flow would work, especially for admin-facing operations.

Here’s my current understanding, please correct me if I’m off:

## Client View Flow
When a user wants to view a term/policy, they’d call something like:
```
GET /api/term-policy/en/term
```
and the response might be:
```json
{
  "message": "Your terms and policies",
  "data": {
    "content": "https://s3.amazonaws.com/yourinfo.html",
    ...
    ...
  }
}
```
At this point the client can perform a `GET https://s3.amazonaws.com/yourinfo.html` and render the stuff however it likes.

## Admin Update Flow
If an admin wants to update an existing entry (say, update the URL to point to a a new version), they’d do something like:
```json
{
  "url": "https://your_new_location/term.html?v=example",
  ...
}
```
I assume in practice this would rarely happen... unless we want to fix a typo or hot-patch an error in the content? mh..

---

## Admin Create Flow

If an admin wants to publish a new version of the term or policy, would they:
1. Manually edit and upload the new HTML file externally (through an S3 console, aws-cli or internal CMS)?
2. And then call:
```
# POST /api/term-policy/en/term
{
    "url": "https://whateverendpoint/new.html?v=example",
    "country": "UK",
    "language":"en",
    "version":"2"
}
```

Am I getting this right ?

andrechristikan:
Hi @Gzerox,

Great work on this feature! I'd like to share some thoughts on `Client View Flow`

### GET /list
Your approach looks good here

### POST /create
I'm thinking we might want to reconsider this endpoint. 
Since we'll be seeding all term policy types into the database as version 1 when the project initially runs, the `/create` endpoint might not be necessary anymore.

### DELETE /delete
Similarly, I believe we might not need this endpoint either, considering our versioning strategy.

### PATCH/PUT /update
Instead of having admins upload files directly or use AWS CLI, what if we create a more streamlined flow? Here's what I'm thinking:

1. Add an endpoint for uploading HTML files to AWS S3
2. When uploading, rename files to `${type}-${version}.html` format
3. Use the existing replace system and presigned URL approach
4. After successful upload, pass the AWS URL/key to the update endpoint

The request DTO could look something like:
```json
{
  "key": "https://aws-s3.com/...",
  "size": 20,
  // other fields...
}
```

For reference, check out the `/profile/update-photo` and `/profile/upload-photo` APIs.
They demonstrate this pattern well and might help clarify what I'm suggesting.

What are your thoughts on this approach?

src/modules/auth/dtos/request/auth.sign-up.request.dto.ts
andrechristikan: Oh my bad.
you're right, I think I mixed it up with my previous `ValidationPipe` settings.

src/modules/auth/controllers/auth.public.controller.ts
andrechristikan: Oh, I didn't notice you changed the type in `user.service.ts`.

Should we create one more DTO then? 
Maybe `UserSignUpRequestDto` to make it clearer which module it belongs to.

src/configs/app.config.ts
```diff
@@ -10,6 +10,12 @@ export default registerAs(
         version,
         globalPrefix: '/api',
 
+        country:{
+            available: process.env.APP_COUNTRY,
+            default: process.env.APP_COUNTRY,
+            header: "x-custom-lang"
```
andrechristikan: I think there might be a mistake here.
Maybe this should be `x-custom-country`?

src/modules/term-policy/validations/require-user-consent.validation.ts
```diff
@@ -0,0 +1,10 @@
+import { ValidationOptions } from 'class-validator';
+import { RequireTrue } from '@common/request/validations/request.require-boolean.validation';
+
+export function RequireUserAgreement(validationOptions?: ValidationOptions) {
+    return RequireTrue({
+        //FIXME: The message override specified here, is not currently being picked up by the current AppValidationFilter
```
andrechristikan: I will check it later.

Fyi
This is already handled by `AppValidationFilter` actually.
You just need to know the name of your validation.

For example, we have `request.is-password.validation.ts`.
This validation is named `IsPassword`:

```ts
export function IsPassword(validationOptions?: ValidationOptions) {
    return function (object: Record<string, any>, propertyName: string): void {
        registerDecorator({
            name: 'IsPassword',
            target: object.constructor,
            propertyName: propertyName,
            options: validationOptions,
            constraints: [],
            validator: IsPasswordConstraint,
        });
    };
}
```

When it returns a validation error, it will return a `messagePath` like `request.isPassword`.
Then on our side, we just need to add the message in `request.json`.

andrechristikan:
So far so good.
I think we're close to finishing this feature.

By the way, I'll be able to do a full review by this weekend.

src/modules/term-policy/validations/require-user-consent.validation.ts
Gzerox: I created the following new decorators (`src/common/request/validations/request.require-boolean.validation.ts`):
1. RequireBoolean (with render message-key: `requireBooleanConstraint`)

on top of it I created:
1. RequireTrue
2. RequireFalse

On top of `RequireTrue` I created `RequireUserAgreement` (which render message-key: `termPolicy.error.requireAgreement` ).
However when a field decorated with `RequireUserAgreement` fails the validations, seems that we render the message key from the `RequireBoolean` instead of `RequireUserAgreement`.

**Im pretty sure this is the intended behaviour, and im the one doing something wrong.**

Gzerox:
I'll mark this ready for review, before it becomes stale and we forget about it :D

Lets see if there is anything to polish before the merge !

andrechristikan:
hi @Gzerox 
its been awhile since last review

i will try to recall it and give a feedback

src/migration/seeds/migration.term-policy.seed.ts
andrechristikan, line 1: documentUrl is supposes to be a string of the content or the url ?

andrechristikan:
there are a few things that need to be adjusted, but dont worry let me do that
anw thanks for ur contribution ❤️

src/migration/seeds/migration.term-policy.seed.ts
Gzerox: You right, that was supposed to be "url" and the seed should contain some example of external link reference to that resource, which could be basicalyl anything, but in our case was supposed to be s3 file reference.

Gzerox:
yeah, very hardly i will reach the quality expected for ready-to-go PR :D

PR #579, opened 2025-06-22
feat(projection): added option to use projection on find

Gzerox:
In my current project I had to add the option to use `projection` in the existing `.findXX` methods.

The existing select option in Mongoose is used to include or exclude specific fields from the query result. 
It supports only simple field-level filtering, like:

```
select: 'name email' // or { name: 1, email: 1 }
```

However, there are scenarios where we need computed or transformed fields returned by the query, such as:

1. Extracting nested elements - first/last item in an array
2. Handling conditional values - such as $ifNull
3. Resolving dynamic fields - such as $getField

By allowing an optional `projection` field in IDatabaseOptions, we give services/repositories the ability to define advanced projections using MongoDB’s syntax, 

From my understanding, this is not possible with just `select`.

---

Decline it or accept if this could be helpful to y'all.

### DIFF
--- a/src/common/database/bases/database.object-id.repository.ts
+++ b/src/common/database/bases/database.object-id.repository.ts
@@ -61,6 +61,10 @@ export class DatabaseObjectIdRepositoryBase<
             }),
         });
 
+        if (options?.projection) {
+            repository.projection(options.projection);
+        }
+
         if (options?.select) {
             repository.select(options.select);
         }
@@ -109,6 +113,10 @@ export class DatabaseObjectIdRepositoryBase<
             }),
         });
 
+        if (options?.projection) {
+            repository.projection(options.projection);
+        }
+
         if (options?.select) {
             repository.select(options.select);
         }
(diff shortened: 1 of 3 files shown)


andrechristikan:
Hi @Gzerox! 👋
Thanks for the PR! I appreciate the effort

I did check ur changes, there are some inconsistencies in it
1. Having both `select` and `projection` might confuse developers - when to use which?
2. Other databases don't support this dual concept, which breaks consistency across different DB implementations

Since we're using Repository Design Pattern, I think `$project` should be used directly within MongoDB aggregation pipelines (raw) rather than as a generic database option. This keeps MongoDB-specific features in the right place and maintains better separation of concerns.

Let me know if these concerns make sense to you. If you agree, feel free to close the PR, but I'm also open to discussing alternative approaches 🙏

Gzerox:
@andrechristikan If the project was supposed to be mongo only, I would have a better case :D.

I keep forgetting that the project is not meant to be mongo-specific.
Make sense what you said, im closing the PR and switching my usage to aggregate !

Thank you for your time !

PR #586, opened 2025-07-04
fix(verification): can not get latest email & mobile number by error field expired

vfiee:
When verifying the email address and mobile phone number, the latest email address or mobile phone number record cannot be obtained. After investigation, it was found that the wrong field expired was used in findOneLatestMobileNumberByUser and findOneLatestEmailByUser methods, and the field in the database was expiredDate


Gzerox:
oh, seems we were looking up non-existing field :D.

Good Catch ! @vfiee

PR #595, opened 2025-07-27
fix(term-policies) - A few fixes & small adjustments

Gzerox:
This PR adds a small fixes and some minor changes:

1. During the `seed:user` we are now also creating `termPolicyAcceptance` for every user.
2. Small Fix during lookup of `url.language`
3. Small fix on `TermPolicy.deleteMany` method
4. Add a endpoint that allow user to reject a previously accepted terms - this is ideal for policy like: COOKIES, MARKETING, etc...


andrechristikan:
Hi @Gzerox 
This is already implemented on version 8 too

Thanks for the contribution!

PR #592, opened 2025-07-24
feat(token-rotation): allow refreshToken to be rotated on every /refresh hit - implements #591

Gzerox:
Hello !

This PR introduces a first draft implementation for [issue #591](https://github.com/andrechristikan/ack-nestjs-boilerplate/issues/591).

The proposed feature enables automatic refresh token rotation, which: when enabled, allows clients to maintain continuous access without being forced to re-authenticate, as long as they remain active. 
This behavior can significantly improve the user experience, especially for consumer-facing applications.

That said, depending on the sensitivity of the application, this may introduce security trade-offs that should be carefully evaluated. 

More details and rationale are discussed in the linked issue.

Feedback and suggestions are welcome!

PR #596, opened 2025-07-27
feat(session): added JTI claim

Gzerox:
Hello !

With this PR, I’m proposing the addition of a jti (JWT ID) claim to both the accessToken and refreshToken, along with tracking this identifier in the session document. 

The main goal is to allow the system to **track and enforce the latest active token pair for a given session**.

**How it works**

1. A new `jti` is generated each time a token pair is issued (on login or refresh).
2. This `jti` **is embedded** in both tokens and saved in the corresponding session entry.
3. Any route decorated with `@SessionJtiProtected()` will compare the incoming token’s jti with the one stored in the session.
    * If they **match** → the token is current and valid.
    * If they **don’t** → the token is considered stale, and access will be denied.

This mechanism helps detect and block usage of old tokens, adding a layer of protection against token replay attacks. 
Any mismatched jti could also be logged for auditing or anomaly detection.

This implementation is compatible with standard token issuance but may require adjustments if we decide to proceed with the optional [refresh token rotation](https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/592).


andrechristikan:
Hi @Gzerox 
This is already implemented on version 8

Anw thanks for the contribution!

PR #607, opened 2025-12-16
feat: add 2fa

### LINKED ISSUE #502: Request for Two-Factor Authentication (2FA) Feature Implementation

ak2g:
I have added 2fa compatible with google and Microsoft Authenticator app. backup codes can be generated and deleted after they are used


andrerektstudios (commented):
Hi @ak2g,

Awesome!
I will review it and give you my feedback. I'll get back to you soon.

Anyway, thank you for contributing!

andrerektstudios (commented):
...

andrechristikan (commented):
Opss, wrong account 😅 haha

package.json
```diff
@@ -145,6 +146,7 @@
         "husky": "^9.1.7",
         "jest": "^30.2.0",
         "lint-staged": "^16.2.7",
+        "otplib": "^12.0.1",
```
andrechristikan, line 149: i dont think need to put this under `devDependencies` because we already add it under `dependencies`

andrechristikan:
Based on the code and doc that u wrote in this PR
There are some point to be mark

## The Feat
i think we need to make this feat to be required from beginning
and also dont need to make this dynamic

## Cache
we need to use default cache for this feat, currently u use session cache provider
`CacheMainProvider` thi is the key

## IV for encryption
i noticed that in this PR `iv` is same for each user

for security reasons i would suggest to make the iv value be different for each user and store it to the db
so in db, we will store 2 important value
1. iv
2. secret

those are will be reset each user active / reactive 2fa token

## Envs
there are to many envs that should not dynamic i think

for example these envs, we can move it to the configs
- AUTH_TWO_FACTOR_DIGITS
- AUTH_TWO_FACTOR_STEP
- AUTH_TWO_FACTOR_WINDOW
- AUTH_TWO_FACTOR_SECRET_LENGTH
- AUTH_TWO_FACTOR_CHALLENGE_TTL_MS
- AUTH_TWO_FACTOR_CACHE_PREFIX_KEY
- AUTH_TWO_FACTOR_BACKUP_CODES_COUNT
- AUTH_TWO_FACTOR_BACKUP_CODES_LENGTH

ak2g:
looking forward on it. i'm enhancing 2fa. adding more endpoints give me sometime.

src/modules/session/session.module.ts
```diff
@@ -29,7 +29,7 @@ import { RedisClientCachedProvider } from '@common/redis/constants/redis.constan
             },
         }),
     ],
-    exports: [SessionUtil, SessionService, SessionRepository],
+    exports: [SessionUtil, SessionService, SessionRepository, SessionCacheProvider],
```
andrechristikan, line 32: dont need to export this because we can use `CacheMainProvider`

src/modules/auth/services/auth-two-factor.service.ts
andrechristikan, line 1: can we change this to util instead of service ?

src/modules/user/dtos/response/user.login.response.dto.ts
andrechristikan, line 1: better to extends PartialType(UserTokenResponseDto) and put twoFactor related into 1 object ?

how do u think ?

src/modules/user/dtos/request/user.two-factor-verify-login.request.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    @ApiProperty({
+        description: '6 digit code from authenticator app',
+        example: '654321',
+        required: false,
+    })
+    @ValidateIf(o => !o.backupCode)
+    @IsString()
+    @IsNotEmpty()
+    @Matches(/^[0-9]{6,8}$/)
```
andrechristikan, line 27: this length will be follow the config
i think it worth it to create a custom validation for this, maybe called `IsTwoFactorCode` or `TwoFactorCode`

src/modules/user/controllers/user.user.controller.ts
andrechristikan, line 1: i  think this can be use by admin also, no ?
and because of that i think the endpoints should be under shared.controller

src/modules/user/dtos/request/user.two-factor-regenerate.request.dto.ts
andrechristikan, line 1: why this is empty ?

src/modules/user/repositories/user.repository.ts
```diff
@@ -189,6 +192,128 @@ export class UserRepository {
...
+                secret,
+                backupCodes: [],
+                enabled: false,
+                confirmedAt: null,
+                updatedBy: userId,
+            },
+        });
+    }
+
+    async confirmTwoFactor(
```
andrechristikan, line 306: I noticed `confirmTwoFactor` is to enable the two factor
how to change that to `enableTwoFactor`? Just to be clearer

src/modules/user/services/user.service.ts
```diff
@@ -1058,39 +1048,16 @@ export class UserService implements IUserService {
                 promises.push(this.userRepository.verify(user.id, requestLog));
```
andrechristikan, line 1105: I think we can use direct execution here without Promise.all

andrechristikan:
hi @ak2g 
edited my comment, pls take a look again

src/modules/user/controllers/user.public.controller.ts
```diff
@@ -115,6 +117,22 @@ export class UserPublicController {
         );
     }
 
+    @UserPublicLoginTwoFactorVerifyDoc()
+    @Response('user.loginTwoFactor')
+    @ApiKeyProtected()
+    @HttpCode(HttpStatus.OK)
+    @Post('/login/2fa')
```
andrechristikan: can we use term verityTwoFactor instead of login 2fa in here ?

ak2g: sure

andrechristikan:
Love ur work, thats clean and well structured

Btw the only left is just user.service, I will continue to review it tonight or tomorrow

src/configs/auth.config.ts
```diff
@@ -49,6 +49,23 @@ export interface IConfigAuth {
...
+        window: number;
+        secretLength: number;
+        challengeTtlInMs: number;
+        cachePrefixKey: string;
+        backupCodes: {
+            count: number;
+            length: number;
+        };
+        encryption: {
+            key?: string;
```
andrechristikan: is it required, no?

src/modules/user/controllers/user.admin.controller.ts
```diff
@@ -209,6 +210,34 @@ export class UserAdminController {
         );
     }
 
+    @UserAdminDisableTwoFactorDoc()
+    @Response('user.twoFactor.disable')
+    @PolicyAbilityProtected({
```
andrechristikan: need add @ActivityLog here

src/modules/user/dtos/user.dto.ts
```diff
@@ -34,6 +34,13 @@ export class UserDto extends DatabaseDto {
     })
     isVerified: boolean;
 
+    @ApiProperty({
```
andrechristikan: good catch! 
i think we need this to quick check instead of `enable` field in twoFactor table

andrechristikan: dont forget to add it into user table

andrechristikan:
hi @ak2g ,
let me know if ur done with all the changes or need any help 

so i can merge it into other branch first, then review and make adjustments 
lastly, merge it to the main

ak2g:
I'm done for this time, I'm working on notification module now.

andrechristikan:
ok, leave the rest to me
and thank for the contribution and the hard work 🔥

ak2g:
This project saved me a lot of time—happy to contribute!
is there any specific reason that admin endpoint does not have url to delete user?

andrechristikan:
Ah yeah, at first I just wanted to keep it simple.
As long as the project works and I didn't have much time to make a full service anyway.

And yeah, if you're asking why, I don't really have a specific reason.
But if you think it's worth adding, then we can add it in the next version.

PR #610, opened 2025-12-25
fix: very small fixes here and there

Gzerox:
Not Much to say, I was going around testing a few stuff and found a few small adjustments.

Nice work with the v8 !


andrechristikan:
Hi @Gzerox 
Ahh, great catch!

Thanks for the correction, and let me merge it

PR #611, opened 2025-12-25
fix(pagination): page evaluation

Gzerox:
fixes scenario in which the user does not provide `page` and/or `perPage` parameters via url query.

In such scenarios the check was returning `NaN` instead of the intended default value, this breaks then the query which complains about the missing "limit/take" parameter,


andrechristikan:
Make sense, but i think it could still be improved a bit.
No worry, let me handle it

PR #612, opened 2025-12-25
fix(forgot-password): properly handle token usage + session revoke

Gzerox:
### Properly invalidates the forgot-password token
When a user resets their password via the `forgot-password` flow, the token was not marked as used, leaving it reusable.

Even though `expiredInMinutes` is intentionally short, the token must be single-use to avoid replay risks.

To address this, I extended the existing `repository.changePassword` method to also handle this scenario.
I’m open to feedback on whether a dedicated `repository.resetPassword` method would be preferable instead.

### Avoids duplicate session revocation logic
Session revocation was previously handled in two different places, executed in the same `Promise.all`, which led to redundant logic and inconsistent behavior.

```javascript
// From user.service.ts > resetPassword
// First place
  this.userRepository.changePassword(
      resetPassword.userId,
      password,
      requestLog,
  ),
  this.sessionUtil.deleteAllLogins(
      resetPassword.userId,
      sessions
  ),
// Second place
  this.sessionRepository.revokeAll(
      resetPassword.userId,
      requestLog
  ),
```

The `userRepository.changePassword` method already takes care of revoking sessions:
```javascript
sessions: {
    updateMany: {
        # NOTE: we are not updating `revokedAt`
        where: { isRevoked: false },
        data: { isRevoked: true },
}
```
As a result, the subsequent call to `sessionRepository.revokeAll` ends up being a `no-op`, since no active sessions remain (isRevoked is already true).

This change centralizes session revocation in `sessionRepository.revokeAll`, ensuring it happens in a single, well-defined place.

**Note: this still leaves a small risk window, as changePassword and revokeAll are not part of the same transaction.**

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -57,7 +57,8 @@
         "VJMV",
         "Vnwnyr",
         "weakpassword",
-        "Zgoh"
+        "Zgoh",
+        "passwordHasheds"
     ],
     "ignorePaths": [
         "node_modules/**",
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -94,7 +94,7 @@ services:
         environment:
             - REDIS_HOST=redis
             - REDIS_PORT=6379
-            - BULL_PREFIX=queue
+            - BULL_PREFIX=Queue
             - USER_LOGIN=admin
             - USER_PASSWORD=admin123
             - REDIS_DB=1
--- a/docs/doc.md
+++ b/docs/doc.md
@@ -239,16 +239,16 @@ Documents paginated response with automatic pagination parameters.
 **Auto-includes:**
 
 - Standard pagination query parameters (depends on type):
-  - **Offset type (default)**:
-    - `perPage` - Data per page (max: 100)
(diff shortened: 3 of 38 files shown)


andrechristikan:
Hi @Gzerox 

First, the reset password flow.
You're right, within that flow` the token should be marked as used`.
We should use a different repository method from `changePassword`, maybe we could name it `resetPassword`.
And I prefer to separate resetPassword and changePassword into different methods rather than combine them into one.

Second, the duplicate session revocation.
Ah yes, you're right again. If I recall correctly, I just added the `revokeAllByAdmin` and `revokeAll` methods so they could be used anywhere.
But after thinking about it again, it seems like it would be better if it was executed together with the others in a database transaction.
So on my end, I will remove and clean up both of those methods.

Could you update the PR ? and don't forget to pull the latest.
Thanks

andrechristikan:
hi @Gzerox 

thats awesome!
let me merge this pr to the development branch

thanks for the contribution

PR #613, opened 2025-12-25
Custom roles

ak2g:
We often need custom roles, and I find this approach works best in the current context.


andrechristikan:
Hi @ak2g 
Thank you for your contribution.

However, this boilerplate already supports custom roles. You just need to create a new role with different permissions through the role module, as documented in `/docs/authorization.md`.

The idea is to keep everything centralized and maintainable.

I believe this approach won't work well with this boilerplate and goes against the main design.

Here are my concerns:
- Custom permissions per user makes maintenance extremely difficult. This will become a problem sooner or later.
- Violates RBAC best practices

PR #616, opened 2025-12-28
fix(term-policy): json override

Gzerox:
## The issue:
Currently whenever we invoke the `repository.accept()`, we are overriding the content of `termPolicy` via:

```javascript
this.databaseService.user.update({
    //....
    data: {
        termPolicy: { // there we go
            [type]: true,
        },
  }
})
```

this leads to update on `termPolicy` like follows:
```
# Before Update:
{"cookies": false, "marketing": false, "privacy": true, "termsOfService": true}
# After Update (marketing accept)
{"marketing": true }
```

Looking around, looks like partial JSON field update, is not supported: 
1. https://github.com/prisma/prisma/discussions/3070
2. https://github.com/prisma/prisma/issues/5057
---
I don't like at-all the current fixes I made, however I did not found a prettier way, open to suggestions.
This workaround to me feels bad, it adds a new DB interactions, its not wrapped in a transactions, and its clearly a workaround for a not-supported feature.

### DIFF
--- a/src/modules/term-policy/repositories/term-policy.repository.ts
+++ b/src/modules/term-policy/repositories/term-policy.repository.ts
@@ -20,7 +20,7 @@ import {
     EnumTermPolicyType,
     EnumUserStatus,
     Prisma,
-    TermPolicy,
+    TermPolicy, UserTermPolicy,
 } from '@prisma/client';
 
 @Injectable()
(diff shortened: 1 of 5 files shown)


src/modules/term-policy/repositories/term-policy.repository.ts
```diff
@@ -145,12 +145,17 @@ export class TermPolicyRepository {
         { ipAddress, userAgent }: IRequestLog
     ): Promise<ITermPolicyUserAcceptance> {
         const acceptedAt = this.helperService.dateCreate();
+        const user = await this.databaseService.user.findUniqueOrThrow({
```
andrechristikan: in the controller, u have already use @UserProtected
means, u have user data inside the request instance (req.___user)

so instead of fetch user from the db here (cause problem n+1)
why dont we use @UserCurrent decorator to get the user data, and then pass the user data to the repository ?

andrechristikan:
hi @Gzerox 

i think there's a limitation with prisma and mongodb.
prisma can't update a specific field, it updates the entire json object instead.
that's weird behavior for mongodb

anw you're right though, good catch!
i left some feedback, let me know what you think?

src/modules/term-policy/repositories/term-policy.repository.ts
Gzerox: lets see if looks better now or if anything else is needed.

andrechristikan:
hi @Gzerox 
cool, let me merge it

thanks

PR #617, opened 2025-12-29
Push notifications with firebase

ak2g:
Activity log is working great for audit log so have to create new module. it uses firebase for push notification.


andrechristikan:
hi @ak2g,
let me merge into feat/notification branch and later i will check it

thanks for the contribution btw

PR #624, opened 2026-02-09
feat!: add organization-based multi-tenancy

Gzerox:
_NOTE: This PR description was partially generated with AI and then manually validated._

## Summary

This PR introduces the first complete tenant/project authorization foundation for multi-tenant SaaS use cases, plus documentation for both modules:

`docs/tenant.md`
`docs/project.md`

The implementation is intentionally scoped as a solid baseline that is safe to extend, not a full authorization architecture rewrite.

## What This PR Introduces

1. Multi-tenant architecture primitives
	1. Tenant context resolution via `x-tenant-id`
	2. Tenant membership model with tenant-scoped roles
	3. Tenant role/permission guards and decorators
	4. Tenant-scoped login endpoint that returns available memberships

2. Project module (tenant-scoped resource layer)
	1. Project CRUD in tenant context
	2. Project member management
	3. Project-scoped permission checks
	4. Clear separation between tenant-level and project-level authorization

3. Documentation
`docs/tenant.md`: architecture, auth flow, decorators, request lifecycle, endpoints
`docs/project.md`: project architecture, layered guards, endpoint contracts, business rules

## Why

This enables practical B2B/SaaS patterns:

- Multiple organizations share one app instance with isolated data boundaries
- A single user can belong to multiple tenants
- Access checks are enforced at the correct scope:
  - Platform scope (admin modules)
  - Tenant scope (tenant resources)
  - Project scope (project resources)

## Scope and Design Intent

This PR intentionally does **not** refactor the existing platform policy module.
Instead, it adds tenant/project-scoped authorization primitives to correctly enforce resource-level checks for tenant/project flows today.

Current authorization paths are intentionally separate:

- `PolicyAbilityProtected` for platform/admin policy checks
- `TenantPermissionProtected` for tenant-scoped checks
- `ProjectPermissionProtected` for project-scoped checks

A follow-up improvement will unify these into one generic, scope-aware authorization layer.

## Key Authorization Behavior

- Tenant checks use tenant member role abilities.
- Project checks use project member role abilities.
- Platform policy checks keep existing behavior (including superadmin bypass in platform policy flow).
- Project routes requiring `:projectId` apply project membership + ability evaluation.
- Tenant routes requiring current tenant rely on `x-tenant-id` + tenant membership validation.

### Tenant

- `POST /api/v1/public/tenant/login/credential`
- `GET /api/v1/shared/tenants/current`
- `PATCH /api/v1/shared/tenants/current/tenant`
[...]

PR #625, opened 2026-02-20
feat(casl-prisma): extended the current CASL implementation

Gzerox:
## Summary

This PR is an attempt to set a foundation for more flexible permission management and stronger CASL support in this codebase.

This branch introduces policy infrastructure changes and an initial activity-log integration to validate the direction.

The main goal of this PR is to:
1. discuss if the proposed path does it even make sense
2. collect feedback and implementation ideas before scaling this pattern to more modules

## What Changed

1. Refactored policy decorator/guard contract
	- `PolicyAbilityProtected` now supports rule-list and structured requirement inputs (`all` / `any`).
	- Added `@PolicyAbilityCurrent()` to expose request-scoped compiled CASL ability.

2. Migrated CASL integration toward Prisma-compatible authorization
	- Added `@casl/prisma`.
	- Policy ability model now supports `effect`, `fields`, `conditions`, `reason`, and `priority`.
	- Added deterministic rule ordering and context placeholder support (e.g. `$userId`).

3. Expanded role ability storage and validation
	- `RoleAbility` schema extended with new CASL-related fields.
	- Role DTOs updated to accept/return these fields.
	- Added condition validator to reject Mongo-style operators and enforce Prisma-style conditions.

4. Updated role/policy guard/service behavior
	- Unified metadata resolution (`getAllAndOverride`) for guard use.
	- Policy service validates requirements and caches compiled ability in `request.__policyAbilities`.
	- Super-admin policy behavior now depends on explicit role abilities.

5. Applied first Prisma query-level authorization flow
	- Service-level filtering now uses `accessibleBy(...)` with compiled ability where implemented.

## First Implementation Scope (Intentional)

A first implementation of this approach has been performed only for:
- `src/modules/activity-log/controllers/activity-log.shared.controller.ts`
- `src/modules/activity-log/docs/activity-log.admin.doc.ts`
- `src/modules/activity-log/services/activity-log.service.ts`

## Feedback Requested

Please focus feedback on:
- decorator API shape and readability
- guard/service responsibility split
- role ability payload ergonomics
- condition syntax constraints and placeholder strategy
- Prisma-level authorization patterns (`accessibleBy`) in service/repository layers


andrechristikan:
Hi @Gzerox 
Thanks for the PR!

I'm a bit tied up at the moment, but I'll review it shortly. 
Appreciate the effort 🚀

Gzerox:
Hey @andrechristikan,
Absolutely no rush !

While playing with CASL implementation, I feel like i have a few doubts and would really appreciate any insights or feedback on.

### Service method intent becomes less explicit
I like `accessibleBy(...)` for enforcing policy semantics, but I’m not fully convinced by a “pure ability-driven” service signature.

Example: `getListCursor` in `src/modules/activity-log/services/activity-log.service.ts`

Original intent is very clear:
```typescript
# This tells the developer immediately: it returns a paginated list which is scoped to a specific user
async getListCursor(
userId: string,
pagination: IPaginationQueryCursorParams,
)
```

If we move to something like:
```typescript
async getListCursor(
pagination: IPaginationQueryCursorParams,
ability: PolicyAbility,
)
```

then the method behavior becomes more implicit:

- the actual data scope depends entirely on the caller’s ability configuration
- tests also become more ability-setup-heavy, because we now need to encode behavior through CASL rules instead of explicit inputs

My concern is not that this is wrong, but that it reduces readability and makes business intent harder to spot.

### Repeated query conditions (accessibleBy + explicit userId) look redundant

In my current PR code, we end up with both:

1. `accessibleBy(ability).ActivityLog`
2. `{ userId }`

You can see a complete case here: https://github.com/Gzerox/ack-nestjs-boilerplate/blob/15be64ba4b763181ec4eb6835a9c3d27fe470c5c/src/modules/activity-log/services/activity-log.service.ts#L53-L59

At first glance, this looks like duplicated filtering (same userId condition effectively enforced twice).

I’m doing that intentionally for `defense-in-depth`:

- `accessibleBy(...)` enforces the authorization policy
- explicit `{ userId }` enforces the endpoint contract (“this endpoint returns only the current user’s records”)

So yes, there is overlap, but the overlap is intended, for mainly 2 reasons:

- if permissions are misconfigured, the explicit owner clamp still prevents cross-user leakage (this is a big no-no)
- the query remains self-documenting (it is obvious this is a self-scoped endpoint)

Gzerox:
### Possible direction for project-level access

I don’t have the full design yet for managing multiple users per resource, but one possible approach:
Assuming we introduce a `Project` entity, we could manage his access via `ProjectMember`:
`ProjectMember(projectId, userId, roleId)`
Where `roleId` points to a predefined/global role definition, for example:

```
  {
    id: "xxxxx",
    name: "project-user",
    description: "Access to project information",
    type: "user",
    abilities: [
      {
        action: ["read"],
        subject: "project",
        conditions: { projectId: "$projectId" }
      }
    ]
  }
```

On each incoming request (or via a scoped cache strategy):

1. Fetch `ProjectMember` entries for the logged-in user.
2. Build the Ability from those entries (resolving scoped placeholders like $projectId)
3. Verify the Ability matches the action/subject/conditions required by the endpoint
4. Apply Prisma-level filtering (accessibleBy) to enforce the same rules at query level

### Limitations

1. Roles are global/predefined only
	We cannot support project-specific custom roles yet. This keeps things simpler, but limits flexibility
2. No per-project policy overrides
	Even if a project needs a slightly stricter/looser policy, we currently can’t express that without creating another global role
3. Cache invalidation / revocation latency
	If we cache ProjectMember entries, role changes or membership removal may not take effect immediately unless we define a clear invalidation strategy.

---
EDIT: While thinking about it, the previous approach I mentioned would require to create a new XXXMember collection for every new model/entity we plan to manage access for.
From a more “framework-oriented” perspective, it might be preferable to adopt a single, centralized collection responsible for managing access across all resource types. This would allow us to extend access control to new entities without requiring additional schema changes.

That said, implementing this via a polymorphic relationship introduces additional complexity, particularly in terms of referential integrity, query ergonomics, and Prisma’s limited support for true polymorphism.

```
model AccessGrant {
  id            String   @id @default(cuid())

  principalType String   // "USER" (future: "TEAM", "SERVICE_ACCOUNT")
  principalId   String   // userId

  resourceType  String   // "PROJECT" | "TENANT" ...
  resourceId    String

  roleId        String
  role          Role     @relation(fields: [roleId], references: [id])

}
```

andrechristikan:
Hi @Gzerox , 
I noticed you created 2 PRs for this project. 

Could you let me know which one should be merged first? 
Based on your answer, I'll start the review and provide feedback accordingly.

Gzerox:
Hey @andrechristikan ,

I think the CASL implementation (this pr) is still missing a few foundational pieces here and there. 
I’d like to spend the next few days playing with it a bit more to make sure we end up with a clearer proposal on how to use the CASL approach, along with documentation that covers multiple scenarios more clearly.

My feeling is that this new approach to managing permissions will be important for supporting more advanced scenarios, such as the tenant / project model.

In the meantime, I think it would make sense to start reviewing the tenant / project PR:  
https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/624

Once the CASL implementation is finalized and ready to be merged, we may need to revisit how permissions are handled in the tenant/project module (and potentially in other modules as well) to align everything with the new permission model.

What do you think ?

Gzerox:
Hey @andrechristikan , long time no see :)
Im back and would like to assist in pushing the repository forward.

Im going to close this particular CASL Proposal, I would like to re-implement this with fresh eyes and more thoughtfully starting with the basics.
Also: Looking at this after a few months, looks to me this PR has too much going on :D

PR #635, opened 2026-03-12
Feat/multi tenant

### LINKED ISSUE #627: Allow service-layer transaction composition across repository calls
The codebase currently places all `$transaction(...)` calls inside individual repository methods. 
While this works for single-repository operations, it creates a limitation: there is no way to wrap multiple repository calls from the service layer in a single atomic transaction. 
This means if a service method calls Repository A and then Repository B, a failure in B leaves the database in a partially-updated, inconsistent state with no rollback of A.

_There are also related DB/Redis split issues (to be discussed separately in another thread), so this issue focuses exclusively on Prisma-level database transactions._

---

Here below a few cases which explains in more detail the concern im referring to:

### UserService.changePassword
Lets take a look at the following snippet https://github.com/andrechristikan/ack-nestjs-boilerplate/blob/development/src/modules/user/services/user.service.ts#L886-L900
```typescript
async changePassword(...){
	//....
	await Promise.all([
	    this.userRepository.changePassword(
	        user.id,
	        password,
	        requestLog
	    ),
	    this.sessionUtil.deleteAllLogins(user.id, sessions),
	    twoFactorVerified
	        ? this.userRepository.verifyTwoFactor(
	              user.id,
	              twoFactorVerified,
	              requestLog
	          )
	        : Promise.resolve(),
	]);
}
```
**What happens today:**
In a `Promise.all`, `changePassword` and `verifyTwoFactor` are called, each with their own internal `$transaction`.

**Why it's a problem:**
- `changePassword` and `verifyTwoFactor` are two separate database transactions running in parallel. 
If `verifyTwoFactor` fails after `changePassword` commits, the password is changed but the 2FA backup code is not consumed.
If a backup code was used, the same backup code can be replayed: it was never consumed from the stored list. 

**What it should be:**
All writes: change password, consume the 2FA backup code, should be a single atomic transaction. 
If any step fails, none of them commit.

### UserService.resetPassword
Lets take a look at the following snippet: https://github.com/andrechristikan/ack-nestjs-boilerplate/blob/development/src/modules/user/services/user.service.ts#L1442-L1475

```typescript
async resetPassword(...){
      await Promise.all([
          this.userRepository.resetPassword(
              resetPassword.userId,
              resetPassword.id,
              password,
              requestLog
          ),
          this.sessionUtil.deleteAllLogins(
              resetPassword.userId,
              sessions
          ),
          twoFactorVerified
              ? this.userRepository.verifyTwoFactor(
                    resetPassword.userId,
                    twoFactorVerified,
                    requestLog
                )
              : Promise.resolve(),
      ]);
}
```
**What happens today:**
`resetPassword` (which invalidates the forgot-password token and sets the new password) and `verifyTwoFactor` (which consumes a 2FA backup code) are two separate repository calls inside a `Promise.all`: each has its own internal `$transaction`.

**Why it's a problem:**
If `resetPassword` succeeds (password updated, forgot-password record marked as used) but `verifyTwoFactor` fails, the backup code it not removed from the user's stored list. 
The password was already reset, but the same backup code remains valid and can be replayed against a different forgot-password token.

**What it should be:**
Both writes must be atomic: the forgot-password token is invalidated and the new password is saved at the same time the 2FA code is consumed. If either step fails, neither commits.

---
**Root Cause**

None of the repository methods accept a caller-supplied `Prisma.TransactionClient`. 
**Every `$transaction` is fully self-contained inside the repository.** 
This makes it architecturally impossible for a service to say "wrap these three repository calls in a single transaction."

```typescript
// Repository method signature change (optional tx, backward compatible):
async changePassword(userId: string, password: IAuthPassword, requestLog: IReqLog, tx?: Prisma.TransactionClient): Promise<IUser> {
    const client = tx ?? this.databaseService;
    // use `client` instead of `this.databaseService`
}

// Service layer controls the transaction boundary:
await this.databaseService.$transaction(async (tx) => {
    await this.userRepository.resetPasswordAttempt(userId, tx);
    await this.userRepository.changePassword(userId, password, requestLog, tx);
    if (twoFactorVerified) {
        await this.userRepository.verifyTwoFactor(userId, twoFactorVerified, requestLog, tx);
    }
});
```

Pros:
- Aligns with Prisma's official recommendation for composable transactions
- Service layer retains full control over transaction scope

Cons:
- Repository method signatures grow by one optional parameter. (IMO: ugly to see)

### Alternative: @nestjs-cls/transactional
Very recently I found out about this module, which look very promising.
Rather than threading an optional `tx?` parameter through every repository method, `@nestjs-cls/transactional` offers a cleaner approach. 
It stores the active transaction in `AsyncLocalStorage`, making it automatically available to any repository in the call chain without changing method signatures.

### Future complexity
As the platform grows, this limitation will become increasingly difficult to work around.

PR #624  which introduces the Tenant and Project modules: already surfaces the same problem at a larger scale. 
Operations like invitation claim and activation, membership state transitions, and role assignments inherently span multiple repositories (UserRepository, TenantRepository, ProjectRepository, MembershipRepository). 
Without the ability to compose a single transaction at the service layer, each of these multi-step flows carries the same partial-update risk described above across more entities and with more ways to land in an inconsistent state.

### LINKED ISSUE #671: [FEAT] Support Nest v12
### Preflight checklist

- [x] I searched existing issues and this is not a duplicate.
- [x] I read the relevant docs in the `/docs` folder.

### Problem Statement

The project is currently pinned to NestJS v11 and its surrounding dependency tree. Staying on v11 has a few costs:

- **We can't adopt v12 features.** ESM-ready packages, [[Standard Schema](https://standardschema.dev/)](https://standardschema.dev/) based validation and serialization, and native [[observability](https://docs.nestjs.com/observability/overview)](https://docs.nestjs.com/observability/overview) are all v12-only. Right now we'd have to hand-roll or bolt on third-party equivalents for anything in that list.
- **Dependency drift compounds.** The longer we wait, the more transitive packages move ahead of what our v11 lock file allows. Ecosystem packages (`@nestjs/*`, ORM drivers, validation and testing libraries) increasingly target v12, so new versions and bug fixes become unavailable to us.
- **Security and maintenance risk.** Older majors receive fewer patches, and out-of-date transitive dependencies are harder to audit and remediate when a CVE lands.
- **A larger, riskier migration later.** Upgrading now, while `nest upgrade` handles most of the mechanical work, is far cheaper than skipping a major and doing a multi-version jump under time pressure.

Upgrading the full dependency set to v12-compatible versions unblocks the new features, keeps us on a supported major, and keeps future upgrades incremental rather than disruptive.

### Proposed Solution

Upgrade all project dependencies to versions compatible with NestJS v12.

NestJS 12 focuses on ESM-ready packages, updated CLI defaults, first-class [Standard Schema](https://standardschema.dev/) support for validation and serialization, and native [observability](https://docs.nestjs.com/observability/overview) support. Our current v11 dependency tree needs to be brought in line with that before we can adopt any of it.

Steps:

Upgrade the Nest CLI globally, since the upgrade command ships with it:
`   npm i -g @nestjs/cli@latest @nestjs/schematics@latest
`Update the local dev dependency copy of the CLI as well:
bash
`   npm i -D @nestjs/cli@latest @nestjs/schematics@latest
`Run the upgrade from the project root:
bash
`   nest upgrade
`

Reference: https://docs.nestjs.com/migration-guide

### Primary affected module

other / unknown

### Entry point

Other / unknown

### Acceptance Criteria

- [ ] Nest CLI and schematics upgraded to latest, both globally and as local dev dependencies
- [ ] `nest upgrade` run from the project root and all `@nestjs/*` packages resolved to v12-compatible versions
- [ ] Remaining third-party dependencies (ORM/driver, validation, serialization, testing, linting) updated to versions compatible with NestJS 12
- [ ] Lock file regenerated and committed; no unmet peer dependency warnings on a clean install
- [ ] `npm run build` completes with no TypeScript errors
- [ ] Full test suite passes (unit + e2e), with no tests skipped or removed to make the upgrade pass
- [ ] Application boots locally and in CI; all modules resolve and health check responds
- [ ] Breaking changes from the [migration guide](https://docs.nestjs.com/migration-guide) addressed and documented in the PR description
- [ ] `npm audit` shows no new high or critical vulnerabilities introduced by the upgrade
- [ ] CI pipeline green on the upgrade branch

### Nice to have

- [ ] Deprecated APIs removed rather than left with compatibility shims
- [ ] README / contributing docs updated with the new minimum Node and CLI versions
- [ ] Staging deployment verified before merge

### API Design (if applicable)

```shell

```

### Database / seed (if applicable)

_No response_

### Environment (if applicable)

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

andrechristikan:
# Multi-Tenant Feature Scope

This document defines the scope for the multi-tenant feature.
Some things are intentionally kept simple for the initial implementation to avoid over-engineering.

> **Scope note:** Everything in this document covers the **user role** only. Admin and super admin roles are out of scope for this initial implementation.

---

## Out of Scope

The following are explicitly **not** included in this initial implementation:

- Admin and super admin endpoints
- JIT (Just-in-Time) access for platform support
- Separate tenant login endpoint — standard login is used with `lastTenantId` in the response
- Standalone user-owned projects — all projects must belong to a tenant

---

## General

- Multi-tenant architecture with two levels: **tenant** and **project**. One tenant can have many projects.
- For enabling/disabling tenancy, use a **feature flag** and let the routes remain registered in code.

---

### Setup
- On sign-up, automatically create a default tenant for the user.
- Tenant name is randomly generated, similar to how we handle usernames.
- Tenant slug is auto-generated from the name on creation. Slug is **globally unique** across all tenants — collision handling applies (random suffix if slug already exists).
- Track the user's last active tenant via `lastTenantId` on the User model. This is set on sign-up, on invite acceptance, and on tenant switch — allowing the frontend to redirect the user to their last active tenant on login or page load.
- Use `x-tenant-id` header to determine the current tenant context, so users can switch between tenants.

### Editable Fields
Only `name` and `description` are editable. No status or archive functionality needed.

`slug` is not updated when `name` changes. It can only be changed via a dedicated endpoint, similar to how username updates work.

### Roles
Tenant roles use an **enum directly** — no relation to the Role model. Roles are static. One tenant has exactly **one owner**.

| Action | Owner | Admin | Member |
|---|---|---|---|
| View tenant | ✅ | ✅ | ✅ |
| Update description | ✅ | ✅ | ❌ |
| Update name | ✅ | ❌ | ❌ |
| Update slug | ✅ | ❌ | ❌ |
| Transfer ownership | ✅ | ❌ | ❌ |
| Manage members (invite, remove, update role) | ✅ | ✅ | ❌ |
| View member list | ✅ | ✅ | ✅ |
| Manage projects (CRUD) | ✅ | ✅ | ❌ |
| View projects | ✅ | ✅ | ✅ |
| Leave tenant | ✅ (with rules) | ✅ | ✅ |
| Delete tenant | ✅ | ❌ | ❌ |

> Tenant `member` can only view projects they are assigned to — not all projects in the tenant.

### Leave & Delete Rules
- `owner` wants to leave → must transfer ownership to another existing member first.
- `owner` is the last member → tenant is immediately soft deleted.

### Transfer Ownership Rules
- Target must be an **existing tenant member**.
- If the intended new owner is not yet a member, invite them first, wait for acceptance, then transfer.

### Soft Delete
- Uses `deletedAt` only. No `status` field.
- When a tenant is soft deleted, within a single transaction:
  - All projects under the tenant are soft deleted.
  - All `pending` invitations for the tenant and its projects are revoked.

---

### Setup
- On sign-up, automatically create a default project under the user's default tenant.
- Project name is randomly generated.
- Project slug is auto-generated from the name on creation. Slug is **unique per tenant** — collision handling applies within the same tenant (random suffix if slug already exists).
- Every project must belong to a tenant — standalone user-owned projects are not supported.
- Deletion uses **soft delete** (`deletedAt`). No `status` field needed.

### Context
Current project is determined by `:projectId` route param — no header needed.
The tenant context is already known from `x-tenant-id`. The guard validates that the project belongs to the current tenant.

```
x-tenant-id: <tenantId>        ← from header
GET /projects/:projectId        ← projectId from param
```

### Editable Fields
Only `name` and `description` are editable.

`slug` is not updated when `name` changes. It can only be changed via a dedicated endpoint, similar to how username updates work.

### Roles
Project roles are `admin`, `member`, `viewer`. Leave this as a **skeleton** — each company will have different requirements.

| Action | Admin | Member | Viewer |
|---|---|---|---|
| View project | ✅ | ✅ | ✅ |
| Update project | ✅ | ❌ | ❌ |
| Update slug | ✅ | ❌ | ❌ |
| Delete project | ✅ | ❌ | ❌ |
| Manage members (invite, remove, update role) | ✅ | ❌ | ❌ |
| View member list | ✅ | ✅ | ✅ |
| Leave project | ✅ (with rules) | ✅ | ✅ |

> Tenant `owner` and `admin` can also invite members to any project — enforced at the guard level, no project `admin` record needed.

### Leave Rules
- Project `admin` wants to leave → must promote another member to `admin` first.
- Project `admin` is the last member → project is immediately soft deleted.

### Member Source & Role Assignment Rules

Only existing tenant members can be invited to a project. There is no direct path from outside the tenant into a project — users must be a tenant member first.

| Who | Tenant membership | Project membership | Assignable project role |
|---|---|---|---|
| Tenant `owner` | Already a member | Auto-member of all projects | `admin` — hardcoded, cannot be changed |
| Tenant `admin` | Already a member | Auto-member of all projects | `admin` — hardcoded, cannot be changed |
| Tenant `member` | Already a member | Can be invited | `viewer`, `member`, `admin` |

> Tenant `owner` and `admin` automatically have project `admin` access across all projects in their tenant.
> This is enforced at the **guard level**, not in the database.

### Revoke Access
Only tenant `owner` and `admin` can revoke any project member's access.

### Soft Delete
- Uses `deletedAt` only. No `status` field.
- When a project is soft deleted, within a single transaction:
  - All `ProjectMember` records are soft deleted.
  - All `pending` invitations for the project are revoked.

---

## Module: Invitation

Invitations are split into two separate contexts: **tenant** and **project**. They are not combined into a single source.

### Invite Type

Before creating an invite, the frontend checks whether the email is already registered via `POST /check/email`. Based on the response, the invite is created with an explicit `type`:

| Type | Condition | Behavior on claim |
|---|---|---|
| `registered` | Email already exists in the system | Auto join — no registration needed |
| `unregistered` | Email not found in the system | Redirected to sign-up using the invite token |

> `type` is set at invite creation time and does not change afterward.
> `type` only applies to **tenant invitation**. Project invitation is restricted to existing tenant members only.

Invited users do **not** get a default tenant and project created — they are joining an existing one.

---

### Tenant Invitation

Who can invite: tenant `owner` and `admin`.

#### Role Rules
- `tenantRole` is required and must be `admin` or `member`.
- `owner` cannot be assigned via invite — ownership can only be transferred through the Transfer Ownership flow.

#### Model
Must include: `invitedById`, `invitedEmail`, `tenantId`, `tenantRole`, `type`, `status`, `expiresAt`, `revokedAt`, `revokedById`.

Member records are **not** pre-created before the invite is accepted.

#### Status Lifecycle
`pending` → `accepted` / `expired` / `revoked`

#### Expiry
Default **7 days**, configurable via config. Inviter can override per request via optional `expiresIn` field.

#### Duplicate Handling
Determined by exact match on `invitedEmail` + `tenantId`:
- Exact match with a `pending` invite → **replace**: existing invite is revoked, new invite is created.

#### Revoke
Tenant `owner` or `admin` can revoke a `pending` invite at any time. Once revoked, the token is immediately invalidated.

#### Notification
Invitee receives an email with an invitation token regardless of registration status.
- **`registered`** → claim with token only. Auto joined to the tenant.
- **`unregistered`** → directed to sign-up using the token. Upon registration, auto joined to the tenant.

#### On Completed
Set `lastTenantId` for the user immediately after joining.

---

### Project Invitation

Who can invite: tenant `owner`, tenant `admin` (via guard), and project `admin`.

> Project invitation is for **existing tenant members only**. Users who are not yet part of the tenant must be invited via tenant invitation first.

#### Role Rules
`projectRole` is required: `admin`, `member`, or `viewer`.

#### Model
Must include: `invitedById`, `invitedEmail`, `projectId`, `projectRole`, `status`, `expiresAt`, `revokedAt`, `revokedById`.

Member records are **not** pre-created before the invite is accepted.

#### Status Lifecycle
`pending` → `accepted` / `expired` / `revoked`

#### Expiry
Default **7 days**, configurable via config. Inviter can override per request via optional `expiresIn` field.

#### Duplicate Handling
Determined by exact match on `invitedEmail` + `projectId`:
- Exact match with a `pending` invite → **replace**: existing invite is revoked, new invite is created.

#### Revoke
Tenant `owner`, tenant `admin`, or project `admin` can revoke a `pending` invite at any time. Once revoked, the token is immediately invalidated.

#### Notification
Invitee receives an email with an invitation token. On claim, they are auto joined to the project with the assigned role.

#### On Completed
No change to `lastTenantId` — user is already a tenant member.

---

## Module: Auth (Login Response)

On login, include a `tenant` object in the response so the frontend can immediately set the correct tenant context without an extra request.

```typescript
{
  accessToken: string,
  refreshToken: string,
  tenant: {
    id: string   // from lastTenantId — always set, default tenant is always created on sign-up
  }
}
```

---

## Config

The following configs are required under `auth.config.ts`:
- `tenant.header` — header key for resolving tenant context (`x-tenant-id`)
- `tenant.invite.defaultExpiresIn` — default expiry for invite tokens, can be overridden per request

---

### 1. Sign-Up Flow

```mermaid
flowchart TD
    A[User Sign Up] --> B[Create User]
    B --> C[Create default Tenant\nrandom name + slug]
    C --> D[Add user as tenant owner]
    D --> E[Create default Project\nrandom name + slug]
    E --> F[Add user as project admin]
    F --> G[Set lastTenantId on User]
    G --> H[Return login response\nwith tenant object]
```

### 2. Login Flow

```mermaid
flowchart TD
    A[User Login] --> B[Validate credentials]
    B --> C[Return access token\n+ tenant object from lastTenantId]
    C --> D[Frontend sets x-tenant-id\nfrom tenant.id]
    D --> E[User lands on last active tenant]
```

### 3. Tenant Invitation Flow

```mermaid
flowchart TD
    A[Owner or Admin\nwants to invite] --> B[Check email via POST /check/email]
    B --> C{Email registered?}
    C -- Yes --> D[Create invite\ntype: registered]
    C -- No --> E[Create invite\ntype: unregistered]
    D --> F[Send email with token]
    E --> F
    F --> G[Invitee clicks email link]
    G --> H{Invite type?}
    H -- registered --> I[Claim with token only\nAuto join tenant]
    H -- unregistered --> J[Sign up with token\n+ password + signUpFrom]
    J --> K[Create User\nAuto join tenant]
    I --> L[Set lastTenantId\nMark invite as accepted]
    K --> L
    L --> M[Done]
```

### 4. Project Invitation Flow

```mermaid
flowchart TD
    A[Owner, Admin, or Project Admin\nwants to invite] --> B[Select existing tenant member]
    B --> C[Create project invite\nwith projectRole]
    C --> D[Send email with token]
    D --> E[Invitee clicks email link]
    E --> F[Claim with token only\nAuto join project with assigned role]
    F --> G[Mark invite as accepted]
    G --> H[Done]
```

### 5. Revoke Tenant Invite Flow

```mermaid
flowchart TD
    A[Owner or Admin\ncalls revoke endpoint] --> B{Invite status is pending?}
    B -- No --> C[Return error\nalready accepted/expired/revoked]
    B -- Yes --> D[Set status to revoked\nSet revokedAt + revokedById]
    D --> E[Token immediately invalidated]
    E --> F[Done]
```

### 6. Revoke Project Invite Flow

```mermaid
flowchart TD
    A[Owner, Admin, or Project Admin\ncalls revoke endpoint] --> B{Invite status is pending?}
    B -- No --> C[Return error\nalready accepted/expired/revoked]
    B -- Yes --> D[Set status to revoked\nSet revokedAt + revokedById]
    D --> E[Token immediately invalidated]
    E --> F[Done]
```

### 7. Tenant Switch Flow

```mermaid
flowchart TD
    A[User switches tenant] --> B[Frontend sets new x-tenant-id]
    B --> C[Request hits tenant guard]
    C --> D[Resolve tenant from x-tenant-id]
    D --> E[Validate user is tenant member]
    E --> F[Update lastTenantId on User]
    F --> G[Continue to controller]
```

### 8. Project Access Flow

```mermaid
flowchart TD
    A[Request with x-tenant-id\n+ projectId param] --> B[Tenant guard\nresolve tenant context]
    B --> C{Is user tenant owner\nor admin?}
    C -- Yes --> D[Grant project admin access\nauto, no DB check needed]
    C -- No --> E[Check ProjectMember record]
    E --> F{Is user project member?}
    F -- Yes --> G[Grant access\nbased on project role]
    F -- No --> H[403 Forbidden]
    D --> I[Continue to controller]
    G --> I
```


andrechristikan:
Hi @Gzerox 

> For more detail on the feature being built, please read the feature scope doc linked above this PR.

Good work on this overall.
That said, there are quite a few things that need to be aligned with the agreed spec before we can merge. 

Grouped by module below, feel free to ping if anything's unclear.

---

## Tenant Module

### 1. Roles — switch to enum, drop the Role model relation

Right now `TenantMember` uses `roleId` pointing to the `Role` model. For tenant roles this is more complexity than we need.

Tenant roles are static (`owner`, `admin`, `member`) and won't change per-tenant. An enum directly on `TenantMember` is cleaner and skips the join on every membership check.

### 2. Login — drop the separate tenant login endpoint

The `/public/tenant/login/credential` endpoint that returns `tenants[]` isn't something we need here.

The plan is to keep standard login and just include `tenant: { id }` (from `lastTenantId`) in the existing login response. Frontend picks that up and sets `x-tenant-id` directly — no extra round trip or tenant selector needed.

### 3. Tenant `status` field — remove it

`EnumTenantStatus` and the `status` field on Tenant aren't needed. `deletedAt` alone is enough to tell if a tenant is active or gone — no need to maintain two separate signals for the same thing.

### 4. A few missing fields on Tenant

- `description` — editable field, needs to be on the model
- `slug` — globally unique, auto-generated from name at creation. Doesn't change when name is updated — slug gets its own dedicated endpoint for that
- `lastTenantId` on User — set on sign-up, on invite acceptance, and on tenant switch. Used by the login response so the frontend knows where to land

### 5. JIT access — remove the whole thing

JIT access adds a significant amount of complexity — temporary membership lifecycle, expiry logic, a dedicated role, admin endpoints, and extra audit trail — that we don't need for this initial implementation. It can always be added later when there's a concrete need for it. For now, remove everything related:

- Role `tenant-platform-support`
- Admin endpoints `assume-access` and `revoke-access`
- Fields `isJit`, `expiresAt`, `revokedAt`, `reason` on `TenantMember`
- Feature flag `tenantJitAccess`
- Activity log entries `tenantJitAccessAssumed` and `tenantJitAccessRevoked`

### 6. Leave & transfer ownership rules need to be added

These behaviors need to be handled:
- `owner` wants to leave → must transfer ownership to another existing member first
- `owner` is the last member → tenant gets soft deleted immediately
- Transfer target must be an existing member — can't transfer to someone who hasn't joined yet

### 7. Soft delete needs a cascade

When a tenant is soft deleted, all of the following should happen in a single transaction:
- All projects under that tenant get soft deleted
- All `pending` invitations for the tenant and its projects get revoked

---

## Project Module

### 1. Remove standalone user-owned projects

The `/shared/projects` controller and `ownerUserId` on Project support projects outside of a tenant context — that's not something we're supporting here. Every project needs to be under a tenant.

Specifically:
- Remove `ownerUserId` from the Project model
- `tenantId` should not be nullable
- Remove the `/shared/projects` controller
- Only `/user/tenants/projects` stays

### 2. A couple of missing fields on Project

- `description` — editable field, same as tenant
- `slug` — unique per tenant (not globally), auto-generated from name at creation. Same pattern as tenant slug — doesn't change on rename, has its own endpoint

### 3. Leave rules need to be added

- Project `admin` wants to leave → must promote another member to `admin` first
- Project `admin` is the last member → project gets soft deleted immediately

### 4. Soft delete needs a cascade

When a project is soft deleted, in a single transaction:
- All `ProjectMember` records get soft deleted
- All `pending` invitations for that project get revoked

---

## Invitation Module

### 1. Split into two separate contexts

The current generic approach (`contextId`, `contextName`, `roleScope`, `inviteType`) works but hides a lot of intent. We'd like this split into two explicit contexts — **tenant invitation** and **project invitation** — each with its own model, role rules, and duplicate logic.

### 2. Use explicit fields per context

**Tenant invitation** model should include: `invitedById`, `invitedEmail`, `tenantId`, `tenantRole`, `type`, `status`, `expiresAt`, `revokedAt`, `revokedById`.

**Project invitation** model should include: `invitedById`, `invitedEmail`, `projectId`, `projectRole`, `status`, `expiresAt`, `revokedAt`, `revokedById`.

Note: project invitation has no `tenantId` or `tenantRole` — the tenant context is already known from the request, and project invites are restricted to existing tenant members only.

### 3. Project invitation is for existing tenant members only

Users outside the tenant cannot be invited directly to a project. They must first join the tenant via a tenant invitation, then be invited to the project. This removes the need for any tenant-related fields or auto-join logic in the project invitation flow.

### 4. Add an invite `type` field — tenant invitation only

Before creating a **tenant** invite, the frontend checks the email via `POST /check/email`. The result determines the invite `type`:
- `registered` — email exists, invitee is auto joined on claim, no registration needed
- `unregistered` — email not found, invitee goes through sign-up using the invite token, auto joined on registration

`type` is locked in at creation time and doesn't change. This field only applies to tenant invitation — project invitation has no `type` since invitees are always existing tenant members.

### 5. Status lifecycle is missing

Add `status` with this flow: `pending` → `accepted` / `expired` / `revoked`.
Also add `revokedAt` and `revokedById` for tracking who revoked and when.

### 6. Default expiry — bump from 5 minutes to 7 days

5 minutes is way too short for an email invite. Default should be **7 days**, configurable via config. Inviters can override it per request with an optional `expiresIn` field.

### 7. Duplicate handling is missing

**Tenant invite:** exact match on `invitedEmail` + `tenantId` with a `pending` invite → revoke the old one, create a new one.

**Project invite:** exact match on `invitedEmail` + `projectId` with a `pending` invite → same, revoke and replace.

### 8. Revoke endpoint is missing

Both invite types need a way to cancel a pending invite:
- **Tenant invite** — tenant `owner` or `admin` can revoke
- **Project invite** — tenant `owner`, tenant `admin`, or project `admin` can revoke

Token should be invalidated immediately on revoke.

### 9. `owner` role must not be assignable via invite

There's no guard against this right now. Ownership transfer has its own dedicated flow — invite `tenantRole` should only accept `admin` or `member`.

### 10. Don't pre-create member records

`TenantMember` and `ProjectMember` records should only be created after the invitee claims or registers with the token — not before sending the invite.

---

## Endpoints

### No `/shared/` endpoints

The scope of this implementation is **user-facing only**. `/shared/` endpoints exist to serve both user and admin contexts — since admin is out of scope here, there's no need for `/shared/`. All endpoints should be under `/user/`.

---

The biggest architectural changes are the **invitation split**, the **role enum vs Role model**, and **removing JIT access** since those affect the overall structure. The rest is mostly filling in missing fields and business rules.

andrechristikan:
@Gzerox 
Do you want to tackle this on your own, or would you be open to collaborating?
We could start by splitting up the tasks if that works for you.

Gzerox:
Hello @andrechristikan ,
Thanks a lot for the detailed review, thats a lot to process ^^.
I think it’s fair that i take care of cleaning this up !

Once everything is aligned, feel free to do the final polish before merging.

I did a first pass on your comments, but i’m currently a bit confused about this part:

> No /shared/ endpoints
> The scope of this implementation is user-facing only. /shared/ endpoints exist to serve both user and admin contexts — since admin is out of scope here, there's no need for /shared/. All endpoints should be under /user/.

After reading this, I’m not fully clear on what your expectations are regarding the endpoints for `tenants`, `projects`, and `invites`, as well as controllers location.

Specifically, I’m struggling to understand how we should handle endpoints that require authenticated users with role-based permissions:  such as `updateSlug`, `inviteMember`, `revokeMember`, etc.
Based on the current structure, all those endpoints should live on `tenants.shared.controller` and `projects.shared.controller`

Without a `/shared` controllers, where would these live?

For example, for the tenants module (controllers under `src/modules/tenant/controllers`), i was thinking of something along these lines:
1. `GET /shared/tenants/<slug>` - logged-in user sees details about specific tenant (if has access to)
2. `GET /shared/tenants` - logged-in user sees all the tenants has access to.
3. `POST /shared/tenants/invite/:token/claim` - user claim the invitation
4. `PATCH/PUT /shared/tenants` - logged-in user updates tenant infos (if has access to)
5. ....
   
   
Similarly, for the projects module (`src/modules/project/controllers`):
1. `GET /shared/projects` - logged in user can see all the project he has access to related to that tenant
2. `GET /shared/projects/xxxx` - logged-in user sees details about specific project
3. `POST /shared/projects/invite/:token/claim` - user claim the project invitation
4. ....


It will take some time to go trough each of your point, but I'll follow up for additional questions if needed.

Thank you in advance!

Gzerox:
@andrechristikan , one more doubt about TenantInvite flow:

> Invitee receives an email with an invitation token regardless of registration status.
> registered → claim with token only. Auto joined to the tenant.
> unregistered → directed to sign-up using the token. Upon registration, auto joined to the tenant.

I’d like to clarify a few edge cases for the unregistered user flow:
1. A non-existing user is invited and receives an email with an “invite accept” link
2. The user clicks the link and is redirected to the signup page
3. The user completes the signup process

At this point, I have a few questions:
1. Email verification
Should the user be considered automatically verified since they accessed the platform through the invite link (which was sent via email)?

2. Tenant join timing
Should the user be automatically added to the tenant immediately after signup, or only after account activation/verification?

3. Flow tracking / context awareness
If we decide to wait for activation, how should we distinguish between:
• a signup coming from an invite link
• a standard user registration

andrechristikan:
Hi @Gzerox 
Let me clarify the structure we're going for.

---

### Role hierarchy

- `super-admin` / `admin` → manage the **platform**
- `user` → owns and manages **tenants and projects**

Tenant-level roles (owner, member, etc.) are just enums scoped within the tenant — not platform-level access.

---

### Endpoint structure

All endpoints in this PR are **user-facing only**. `/shared/` is out of scope here — if only users can access it, it lives under `/user/`.

We'll handle `/admin/*` in a separate PR. Admin and user contexts have fundamentally different logic — admin operates on the platform level (e.g., manage any tenant, audit), while user operates within their own scope. Merging both into the same controllers would break separation of concerns and make the codebase harder to maintain and extend.

---

### Why `:tenantId` vs `x-tenant-id` header?

**Tenant endpoints use `:tenantId` in the URL** because tenant is a first-class resource — you're directly operating *on* the tenant itself (read, update, manage its members).

**Project endpoints use `x-tenant-id` header** because a project always exists *within* a tenant context. The tenant is not the resource being operated on — it's the ambient context. Putting it in the header keeps the URL clean and makes it clear that the tenant is just the scope, not the subject.

#### Example Endpoints

The endpoints below are examples to illustrate the pattern, naming and structure can still be adjusted. 
The key takeaway is the separation of concerns between URL param vs header context.

```
# Tenant context from URL param — tenant IS the resource
GET    /user/tenants
POST   /user/tenants
GET    /user/tenants/:tenantId
PUT    /user/tenants/:tenantId
POST   /user/tenants/:tenantId/members/invite
DELETE /user/tenants/:tenantId/members/:memberId/revoke
```

```
# Tenant context from x-tenant-id header — tenant is the scope, project is the resource
GET    /user/projects
POST   /user/projects
GET    /user/projects/:projectId
PUT    /user/projects/:projectId
POST   /user/projects/:projectId/members/invite
DELETE /user/projects/:projectId/members/:memberId/revoke
```

```
# Public — no auth, no context
POST   /public/tenants/invite/:token/claim
POST   /public/projects/invite/:token/claim
```

JIT is also out of scope for now, we'll remove it from this PR.

---

### Invite flow — unregistered user

#### 1. Email verification

Yes, auto-verified. The invite link is sent via email, so if the user can access it, the email is implicitly valid. No separate verification step needed.

#### 2. Tenant join timing

Since the user is auto-verified on signup (point 1), they also **auto-join the tenant immediately after signup**. No need to wait for a separate activation step.

#### 3. Flow tracking / context awareness

No new signup page or endpoint needed — reuse all existing signup endpoints as-is.

Add an optional `inviteToken` field to `UserSignUpRequestDto` (and the social equivalent):

```ts
@ApiProperty({
    description: 'Optional invite token, required when signing up via invitation link',
    required: false,
})
@IsString()
@IsOptional()
inviteToken?: string;
```

The frontend sources `inviteToken` from the URL query param (`/sign-up?inviteToken=xxx`) and passes it in the request body. The service layer then branches accordingly:

- `inviteToken` present → validate token, auto-verify, auto-join tenant after signup
- `inviteToken` absent → proceed with existing signup flow, no changes to existing behavior

The `from` field stays as-is using existing `EnumUserSignUpFrom` values (`mobile`, `website`) — no new enum value needed.

#### Edge cases

- **Token expired or revoked at signup time** → reject the entire signup. No partial success — we don't want users in a state where they've signed up but have no access to the tenant they were invited to.
- **Token valid** → signup proceeds, user is verified and joined to the tenant in the same transaction.

### Invite Model

For the tenant invite model, on top of what's already defined in the scope (`invitedById`, `invitedEmail`, `tenantId`, `tenantRole`, `type`, `status`, `expiresAt`, `revokedAt`, `revokedById`), we need two additional fields to track claim:

```prisma
acceptedByUserId  String?   @db.ObjectId
acceptedAt        DateTime?
```

Both are optional
only populated when the invite is claimed and status transitions to `accepted`. This covers both the registered and unregistered user flows.


---

Does that align with your understanding? 
Happy to discuss if anything feels off 🙌

Gzerox:
Hey @andrechristikan, I believe we're near completion. I've done refactoring and cleanup per your guidelines,here are a few points I'd like you to verify/advise on:

**Architecture & Design:**

1. Invite expiration handling — Currently, invites don't transition to an `expired` state; I don't think its a real issue since validity is only checked via `expiresAt`. 
Should we add an explicit expiration job, or is lazy validation acceptable from a business logic standpoint?
2. Default tenant/project on signup — When users sign up via `TenantInvite`, we call `signupAndAccept` but don't create a default Tenant+Project. For consistency across both signup flows (regular and invitation-based), should we create these independently regardless of the signup path? ( think yes)
3. Endpoint separation — I'm still uncertain about the endpoint structure user/shared, hope you don't mind taking care of it.

**Code Quality Issues:**

4. Utility duplication — `tenant.util.ts` and `project.util.ts` have overlapping methods. Consider extracting shared logic to a common utility.
5. Slug generation duplication — `createSlug` and `createUniqueSlug` duplicated across modules, also the implementation is not really good, but it works.
6. Revoke method signature — The signature and initial validation checks in `project-member.service.ts:revoke` need refinement.
7. Invite interface — On the Invite module, im left with `src/modules/invite/interfaces/invite.interface.ts`, however its still used for utils/config on Project/Tenant.

Configuration Question:

8. `TenantId` parameter detection — `tenant.service.ts:validateTenantGuard` tries to detect `tenantId` from URL named-params (hardcoded) then header. Should the parameter name be configurable or standardized?

Im sure there are some other small things here and there, this is what comes to my mind as of now.

Let me know if you currently see big gap compared to the desired state:  https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/636

I'try from now on to keep PR way smaller, lesson learned (I hope :D)

andrechristikan (commented):
Hey @Gzerox, great progress!

Here's my take on each point:

**Architecture & Design:**

1. **Invite expiration** — Lazy validation via `expiresAt` is acceptable for now. This approach is also consistent with how we handle email verification and forgot password flows.
2. **Default tenant/project on signup** — Agree. Signup flows should behave consistently. If regular signup creates a default Tenant+Project, invitation-based signup should too — meaning the invited user will have 2 tenants from the moment they sign up.
3. **Endpoint separation (user/shared)** — Feel free to leave it. I'll take care of this, no worries.

**Code Quality:**

1. **Utility duplication** — For now it's fine. I believe `tenant.util.ts` and `project.util.ts` will diverge in purpose over time, and keeping modules separate is the right direction for this project.
2. **Slug generation** — Yeah, I saw the commits — there's a lot of duplicate code. Let's move it into a shared util, but keep separate implementations for tenant and project since they may diverge.
3. **Revoke method signature** — Got it. Feel free to leave it to me if you're stuck on this one.
4. **Invite interface** — Fine where it is for now. We can revisit during a proper module boundary review.

**Configuration:**

1. **`tenantId` param detection** — Let's standardize: `x-tenant-id` for header, `tenantId` for URL params. No need to make it configurable at this stage.

This week I plan to jump into this feature and help out. 
Let me know once you're done and I'll review and polish it from there.

Gzerox:
Hey @andrechristikan 

**Default Tenant & Project on Signup**
When users sign up:
1. Invited users: Tenant + project are created immediately upon completing signup via token
2. Regular users: Tenant + project are created when they activate their account (email verification)

In both cases, the actual tenant/project creation happens during the account activation phase.

---

**Utility Duplication Cleanup**
Consolidated tenant/project creation logic:
1. `src/common/helper/services/helper.service.ts` - Base implementation for tenant/project creation
2. `src/modules/project/utils/project.util.ts` - Delegates to `helper.service.ts`
3. `src/modules/tenant/utils/tenant.util.ts` - Delegates to `helper.service.ts`

---

Hopefully this is enough for you the start the review, let me know if you find anything else.

andrechristikan:
hi @Gzerox, 
sure and thanks, i will start the review soon

mosabalrsaheed:
@andrechristikan is this planned anytime soon?! no rush just asking here.

andrechristikan:
hi @mosabalrsaheed ,
probably end of this month

Gzerox:
Hello @andrechristikan ,
Im planning to work on the next few months on some new stuff, which are based on some of the capabilities of this PR.
I wanted to check with you if there is anything I can do to support you on concluding the review/rework on this PR.

Thanks :)

Gzerox:
unbelievable @andrechristikan , while I was writing the previous comment you just committed a lot of work :D

andrechristikan:
hi @Gzerox
thanks for ur patience 😁

there have been so many changes here, and most of them are already done
just 2 things left:
- migrate to ism
- implement unit test

soon enough i will merge this PR

Gzerox:
@andrechristikan , no need to be sorry, you did outstanding work !

For the testing part, would you be open to handling that separately from this PR?
I currently have some WIP changes around tests: added unit tests, an e2e skeleton, and started experimenting with a Jest -> Vitest migration. 

It’s still far from complete, but it might give you a useful starting point and speed things up.

https://github.com/andrechristikan/ack-nestjs-boilerplate/pull/677

Small note: my test changes were done on top of the `feat/postgresql-migration` branch, so they are not directly based on this PR branch. My bad on that. 
Still, maybe there is something useful there that can be reused or adapted (i did not spent too much time on it - may not fit your quality standard :D)

Feel free to brutally reject/ignore it :)

andrechristikan (commented):
## Main features
- **Multi-workspace tenancy** — every user belongs to a workspace; invites and join requests included
- **Projects** — workspace-scoped projects with membership, roles, and soft-delete
- **Analytics dashboard** — admin and current-workspace metrics, anomaly, and fraud views
- **Activity log flush** — domains stage actions; a global interceptor writes after the request settles
- **NestJS 12 + Domain layer** — ESM stack upgrade with Controller → HTTP → Domain → Repository
- **Doc / OpenAPI kit** — Swagger co-located on runtime decorators; kit-only error responses
- **Vitest unit suite** — Jest replaced; path aliases and analytic coverage added
- **Request Zod schemas** — validation moved onto Zod request schemas

## Details

### Architecture and tooling
Foundation for the rest of the branch: runtime, package manager, and module shape.
- NestJS 12 migration; tooling globs, aliases, knip, and tsconfig aligned
- pnpm and dependency set refreshed; lockfile aligned
- Boot and shutdown exit cleanly on failed startup
- Service interfaces dropped; helpers split into kit services
- Router mounts under `router/http`; processors join via the router

### Workspace and project
Product surface for tenancy and collaboration.
- Mandatory multi-workspace with invites and join requests
- Project module with member-based access and static role guards
- Default workspace and project created on email verification
- Unique-slug generation and soft-delete
- Feature flags for invitations and related flows

### Auth, session, and device
Identity path hardened for the new stack and ownership rules.
- Session revoke order fixed; JWT and cache providers registered where utils resolve them
- Device ownership rooted on refresh writes
- Two-factor verification gaps closed; generated unique values bounded
- Auth/util and session/util DI wiring fixed for Nest 12

### Analytic and activity log
Observability for admins and workspace operators.
- Admin and workspace analytics dashboard endpoints
- Analytic response DTOs typed with unit coverage
- Activity-log writes staged through a global flush interceptor
- Workspace activity-log util merged into the repository layer

### OpenAPI / Swagger
API docs stay next to the runtime decorators.
- Doc kit renamed; OpenAPI co-located on runtime decorators
- Kit-only OpenAPI errors; response entries accumulate in the kit

### Request
HTTP edge behavior for clients.
- Request validation moved onto Zod schemas
- Helmet headers on JSON and file responses
- Rate limiting split into three Redis sliding-window limiters
- Pagination pipes replaced; cursor/offset with raw query passthrough

### Queues, config, and platform
Background work, config units, and shared integrations.
- Every enqueue moved into a queue class; utils kept pure
- Config durations carried in their unit; URL patterns and invite links updated
- Health checks split; third-party indicators separated
- File upload validation and export caps made stricter
- Firebase accepts bare base64 private keys; key normalization on a util

### Testing
Unit suite moved and expanded with the analytic surface.
- Vitest with path aliases for specs
- Analytic HTTP and response DTO coverage added
- Shared `test/setup.ts` and Vitest config

### Docs and Claude harness
Guides and agent harness matched to the Domain layer.
- Module guides, README, and contributing repaired
- File upload and presign docs merged; language message docs renamed
- Claude skills collapsed to four; rules/agents matched to the planner pipeline
- Permissions and hook gates scoped to what a commit touches

andrechristikan (commented):
hi @Gzerox,
thanks, and same to you — you did outstanding work here too
you started the multi-tenant work, i mostly cleaned it up. the foundation was solid
really good work

for testing, yes i'll leave it out of this PR
i'll handle it later together with the postgres migration
for now i'm only implementing the config, scoped to `src/app`

for multi-tenant:
it is done on my side, maybe tomorrow i'll release it as v9
feel free to check it out, it's quite different

thanks again for the patience 😁

andrechristikan (commented):
@Gzerox, btw you have 2 feats (postgres migration and the test)
which one should i review first? probably not soon, just so i know the order

PR #638, opened 2026-03-29
fix(activity-log): losing property metadata-activity-log

Gzerox:
## Summary

Fixes missing `metadataActivityLog` values in activity log records on endpoints that use both `@ActivityLog(...)` and `@Response(...)`.

Services are already returning the correct shape — the data just wasn't reaching the activity logger:

```ts
return {
  data: createdUser,
  metadataActivityLog: { userId: createdUser.id },
};
```

## Problem

Both `@ActivityLog` and `@Response` register **interceptors** — middleware that wraps the request/response cycle. The order they're declared in code determines which one processes the response first.

TypeScript applies method decorators **bottom-to-top**: the bottom decorator is registered first and becomes the **outer** interceptor. Outer interceptors receive the response **last**, after all inner ones have already processed it.

When the decorators were ordered like this:

```ts
@Response(...)      // ← top → registered second → INNER → processes response first
@ActivityLog(...)   // ← bottom → registered first → OUTER → processes response last
```

The flow was:

```
Handler returns { data, metadataActivityLog }
  → ResponseInterceptor (inner) normalizes to { statusCode, message, data }  ← metadataActivityLog stripped
  → ActivityLogInterceptor (outer) reads the response                         ← field already gone
```

`ResponseInterceptor` strips `metadataActivityLog` when building the HTTP response, so by the time `ActivityLogInterceptor` tried to read it, it was already gone.

The action itself was still logged, but without the context metadata (e.g. which user was created, which API key was revoked, we basically loose track of which entity was the audit log related to).

## Reproduction

1. Call any mutation endpoint using both `@ActivityLog` and `@Response` — e.g. `POST /admin/user/create`.
2. Have the service return `metadataActivityLog` alongside `data`.
3. Inspect the stored activity-log record.

**Actual:** the record exists, but `metadataActivityLog` is empty.
**Expected:** the record includes the fields returned by the service.

## Fix

Reordered decorators so `@ActivityLog` is always **above** `@Response`:

```ts
@ActivityLog(...)   // ← top → registered second → INNER → processes response first ✓
@Response(...)      // ← bottom → registered first → OUTER → processes response last
```

Now the flow is:

```
Handler returns { data, metadataActivityLog }
  → ActivityLogInterceptor (inner) reads metadataActivityLog   ✓ still present
  → ResponseInterceptor (outer) normalizes the HTTP response
```

Clients never see `metadataActivityLog`  that hasn't changed, the activity log now stores it correctly.

## Additional: decorator ordering cleanup

`@ActivityLog` and `@Response` are interceptors, not auth guards. Leaving them in the middle of a guard stack made them hard to reason about.

This PR also moves both decorators to the **bottom of the stack**, after all auth guards, just before `@HttpCode` and the HTTP method:

```ts
@SomeDoc()
// auth guards ...
@ApiKeyProtected()
@ActivityLog(...)   // ← interceptors grouped at the bottom
@Response(...)
@HttpCode(...)
@Post('/endpoint')
```

`CLAUDE.md` has been updated to document this as the canonical decorator order.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -72,13 +72,14 @@ module/
 @TermPolicyAcceptanceProtected(...)    // 2. Term policy
 @PolicyAbilityProtected({...})         // 3. CASL policy
 @RoleProtected(...)                    // 4. Role check
-@ActivityLog(...)                      // 5. Activity log
-@UserProtected()                       // 6. User status check
-@AuthJwtAccessProtected()              // 7. JWT validation
-@FeatureFlagProtected(...)             // 8. Feature flag
-@ApiKeyProtected()                     // 9. API key
-@HttpCode(HttpStatus.OK)               // 10. HTTP status (only when needed)
-@Get('/endpoint')                      // 11. HTTP method (always last)
+@UserProtected()                       // 5. User status check
+@AuthJwtAccessProtected()              // 6. JWT validation
+@FeatureFlagProtected(...)             // 7. Feature flag
+@ApiKeyProtected()                     // 8. API key
+@ActivityLog(...)                      // 9. Activity log
+@Response('key')                       // 10. Response / @ResponsePaging
+@HttpCode(HttpStatus.OK)               // 11. HTTP status (only when needed)
+@Get('/endpoint')                      // 12. HTTP method (always last)
 async method() {}
 ```
 
--- a/src/common/response/interfaces/response.interface.ts
+++ b/src/common/response/interfaces/response.interface.ts
@@ -34,7 +34,8 @@ export interface IResponseCacheOptions {
 
 /**
  * Mixin shape for controller return values that include activity log metadata.
- * When present, `ResponseInterceptor` forwards this to the `@ActivityLog` handler.
+ * When combined with `@ActivityLog`, place `@ActivityLog` above `@Response`
+ * so the activity log interceptor can read this field before the response is normalized.
  */
 export interface IResponseActivityLogReturn {
     metadataActivityLog?: IActivityLogMetadata;
--- a/src/modules/api-key/controllers/api-key.admin.controller.ts
+++ b/src/modules/api-key/controllers/api-key.admin.controller.ts
@@ -76,7 +76,6 @@ export class ApiKeyAdminController {
     constructor(private readonly apiKeyService: ApiKeyService) {}
 
     @ApiKeyAdminListDoc()
-    @ResponsePaging('apiKey.list')
     @TermPolicyAcceptanceProtected()
     @PolicyAbilityProtected({
         subject: EnumPolicySubject.apiKey,
@@ -86,6 +85,7 @@ export class ApiKeyAdminController {
     @UserProtected()
     @AuthJwtAccessProtected()
     @ApiKeyProtected()
+    @ResponsePaging('apiKey.list')
     @Get('/list')
     async list(
         @PaginationOffsetQuery({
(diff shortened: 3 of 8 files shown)


andrechristikan:
Hey @Gzerox, 
Thanks for taking the time to investigate and open this PR! 🙌

Your instinct about interceptor ordering is right
But the execution order works the opposite of what's described here.

In NestJS, decorators are applied bottom-to-top, but interceptors execute as a nested pipeline. 
The topmost decorator becomes the outermost interceptor, it runs first on the way in and last on the way out.

So with the current order:

```typescript
@Response(...)     // → outer interceptor
@ActivityLog(...)  // → inner interceptor (closer to the controller)
```

The execution flow is:

```
ResponseInterceptor starts
  └─ ActivityLogInterceptor starts
       └─ controller() returns { data, metadataActivityLog }
            ↑ ActivityLogInterceptor.tap() reads metadataActivityLog ✅
       ↑ ResponseInterceptor.map() transforms response, strips metadataActivityLog from HTTP output ✅
```

Because `ActivityLogInterceptor` is inner, its `.tap()` sees the raw controller return value, including `metadataActivityLog` before `ResponseInterceptor` has a chance to transform it.

If we apply the fix in this PR and move `@ActivityLog` above `@Response`, it becomes the outer interceptor instead. 
Its `.tap()` would then receive the already-transformed response `{ statusCode, message, data }` from `ResponseInterceptor` and `metadataActivityLog` would genuinely be missing at that point. That would actually introduce the bug.

The current ordering is intentional and working correctly. 
Closing this PR for now, but really appreciate the investigation!

Gzerox:
Hey @andrechristikan,
I’m going to push back a bit on this one 😈 . 
Hope you don’t mind, let me explain what I’m seeing.

## Expected Behaviour
Given the current decorator order:
```ts
@Response(...)
@ActivityLog(...)
```

one would expect the following execution flow:
```
ResponseInterceptor.before()
  ActivityLogInterceptor.before()
    handler()
  ActivityLogInterceptor.after()
ResponseInterceptor.after()
```

## Actual Behaviour
After adding simple log statements in both interceptors (before next.handle() and inside tap()), the actual output is:
```
[ActivityLogInterceptor] [FLOW][IN] ActivityLog loginWithCredential /api/v1/public/user/login/credential
[ResponseInterceptor] [FLOW][IN] Response loginWithCredential /api/v1/public/user/login/credential
[ResponseInterceptor] [FLOW][OUT] Response loginWithCredential /api/v1/public/user/login/credential
[ActivityLogInterceptor] [FLOW][OUT] ActivityLog loginWithCredential /api/v1/public/user/login/credential
```

This clearly shows the real execution flow is:
```
ActivityLogInterceptor.before()
  ResponseInterceptor.before()
    handler()
  ResponseInterceptor.after() // Strips out metadataActivityLog
ActivityLogInterceptor.after()
```

## Root cause

This aligns with how NestJS actually works, like you said:

> Decorators are applied **bottom-to-top**, and interceptors are registered in that order and executed as a nested pipeline.

So with:
```ts
@Response(...) // Registered second (inner)
@ActivityLog(...) // Registered first (outer)
```

Results in:
```
Inbound: ActivityLog -> Response -> handler
Outbound: handler -> Response -> ActivityLog
```

## After Fix

With the proposed code-changes, the logs become:
```
[ResponseInterceptor] [FLOW][IN] Response loginWithCredential /api/v1/public/user/login/credential
[ActivityLogInterceptor] [FLOW][IN] ActivityLog loginWithCredential /api/v1/public/user/login/credential
[ActivityLogInterceptor] [FLOW][OUT] ActivityLog loginWithCredential /api/v1/public/user/login/credential
[ResponseInterceptor] [FLOW][OUT] Response loginWithCredential /api/v1/public/user/login/credential
```

Which matches the intended flow:
```
ResponseInterceptor.before()
	ActivityLogInterceptor.before()
    	handler()
	ActivityLogInterceptor.after()
ResponseInterceptor.after()
```

In this way we don't lose `metadataActivityLog` and the `ActivityLog Interceptor` properly logs the metadata !

---

I hope you would reconsider reopening this PR, let me know if you are still not convinced, I'll try other approaches 😄

Gzerox:
@andrechristikan sry for pushing again, just wanted to make sure this does not get lost :)

andrechristikan:
Hi @Gzerox

No problem, and yes, I remember it. 
I know the last PR didn't fully solve it.

Let me walk you through my plan. I want to keep the activity log and the response decoupled, because I don't want the activity log to depend on the response decorator.

To do that, I'll implement `nestjs-cls` (AsyncLocalStorage) to set and get metadata across the pipe, service, controller, and any other module.

This way we don't have to worry about decorator order anymore. It's also useful for the tenant PR: with this in place, we no longer need to figure out "which tenant am I currently in?"

The PR will be pushed soon.

PR #646, opened 2026-05-17
fix: remove weak default encryption keys from .env.example

saaa99999999:
## Summary

The  file contains weak, predictable default values for two critical encryption keys that pass all validation checks and would be accepted at startup.

### 1. APP_ENCRYPTION_SECRET_KEY (AES-256 password encryption)

 (BEFORE):

This key is exactly 32 characters, passing the  validation in :

The key flows into AES-256 encryption via :

Used as the encryption key in :

This encrypts user passwords in  via .

### 2. AUTH_TWO_FACTOR_ENCRYPTION_KEY (TOTP secret encryption)

 (BEFORE):

Passes  validation in :

Used for TOTP secret encryption in :

## Fix

Change both keys to empty strings so class-validator rejects them at startup:

 (AFTER):

 (AFTER):

With empty values,  will fail on the encryption key and  will fail on the TOTP key, causing  to throw  and prevent startup until proper random keys are set.

## Impact
- CWE-1188: Insecure Default Initialization
- Predictable encryption keys in example config undermine AES-256 encrypted password and TOTP secret storage
- Developer who copies .env.example without changing values deploys with publicly-known keys


saaa99999999:
## CVE Request — Action Needed from Maintainer

This PR fixes security vulnerabilities. To assign a CVE number:

**GitHub only issues CVEs from the official upstream repository**, not from forks.

Please:
1. Go to this repo → **Security** → **Advisories** → **New draft security advisory**
2. Add `@saaa99999999` as a **collaborator**
3. I will populate the full vulnerability details (CVSS, CWE, data flow, PoC) and submit the CVE request

If you prefer, I can submit the CVE via MITRE (cveform.mitre.org) instead — just let me know.

Thank you for reviewing this PR!

Gzerox:
This PR is clearly generated by some kind of automation without even human double-checking before opening it, missing evidence, missing screenshot, missing code reference.

Im kind enough to reply anyway: The values you are referring to are clearly supposed to be a "example" and should be considered as "quick start", the whole point of the repository is to have a quick and solid starting point.

The only thing I can think of is to be a little bit more clear on the docs: https://github.com/andrechristikan/ack-nestjs-boilerplate/blob/c1009e8ed004f8b439f0868c34081443ddfc73d1/docs/environment.md?plain=1#L197-L200 

explicitly suggesting those values should be randomized/changed by the developers.

saaa99999999:
Thanks for the honest feedback, and for taking the time to reply.

I understand why the CVE request comment looks templated — it follows a standard format to be clear about the process. But the actual vulnerability report in the PR body is based on manual code audit. Each finding traces a specific file path and line number (e.g., .env.example showing the key flow through IsStrongPassword() validation into AES-256 encryption in password.helper.ts). I read the source code to find these.

Your point about documentation is fair — those values are intended as examples for quick start. The concern is that a developer copying .env.example gets predictable encryption keys without any warning at startup. The PR makes both keys empty so the existing class-validator checks catch them, forcing the developer to set proper values before the app runs. That way the documentation and the code tell the same story.

I'm happy to adjust the PR if you have a different approach in mind. The docs update you linked to would also work as a complementary fix.

andrechristikan:
Hi @saaa99999999
Thanks for the report, this is already resolved in development as per your report.

Both APP_ENCRYPTION_SECRET_KEY and AUTH_TWO_FACTOR_ENCRYPTION_KEY are now empty by default, so class-validator rejects them at startup and forces developers to set their own keys.

I'll also clarify in `docs/environment.md` that these values must be randomized per deployment, as suggested above by @Gzerox
I'm closing this PR as the fix is already in place.

PR #670, opened 2026-08-30
fix: quote skill frontmatter YAML; update stale path in rules/database.md

bronsonhill:
Two `SKILL.md` `description:` scalars contain `: ` mid-scalar, which some YAML parsers reject. the frontmatter then yields no `name`/`description`, so anything reading it (skill loaders included) sees two undescribed skills. 

It has been rewritten as `>-` folded block scalars. Also updates `rules/database.md` to stop citing `buildDatabaseExtension` / `database.function.constant.ts`, deleted in the move to `DatabaseExtensionUtil`.

No behavior change intended anywhere.


bronsonhill:
bump @andrechristikan

andrechristikan:
hi @bronsonhill
good catch and thanks
also sorry for the slow response on this one

i verified both parts:
- `migration-seed` and `pr-doc` frontmatter did fail strict YAML parsing because of the mid-scalar. With your `>-` rewrite both parse cleanly, and the description text is byte-identical to the original

- `buildDatabaseExtension` / `database.function.constant.ts` are indeed gone. The updated `rules/database.md` wording matches the current `DatabaseExtensionUtil` setup

fyi, the same issue also exists in `.claude/agents/coder.md` and `.claude/agents/pr-doc-writer.md`
i will fix those separately so this PR can go in as-is

